PR #2, opened 2024-04-10
Optimism RPC Types

EmperorOrokuSaki:
## Motivation

Updating the repository with Optimism RPC types.

## Solution

Implement the needed OP specific RPC types and re-export whenever feasible.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/op-rpc-types/src/op/transaction/tx_type.rs
```diff
@@ -0,0 +1,54 @@
...
+// Identifier for an Optimism deposit transaction
+pub const DEPOSIT_TX_TYPE_ID: u8 = 126;
+
+/// Transaction Type
+///
+/// Currently being used as 2-bit type when encoding it to [`Compact`] on
+/// [`crate::TransactionSignedNoHash`]. Adding more transaction types will break the codec and
+/// database format.
+///
+/// Other required changes when adding a new type can be seen on [PR#3953](https://github.com/paradigmxyz/reth/pull/3953/files).
```
mattsse, line 26: this can be removed, is outdated

crates/op-rpc-types/src/op/block.rs
```diff
@@ -0,0 +1,1516 @@
...
+    pub fn into_full_block(self, txs: Vec<Transaction>) -> Self {
+        Self { transactions: BlockTransactions::Full(txs), ..self }
+    }
+}
+
+/// Block header representation.
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[derive(Default, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct Header {
```
mattsse, line 57: is this the same header as ethereum?

EmperorOrokuSaki: Yes. Optimism RPC's return the same data.

crates/op-rpc-types/src/op/transaction/receipt.rs
```diff
@@ -0,0 +1,196 @@
...
+///
+/// This type is generic over an inner [`ReceiptEnvelope`] which contains
+/// consensus data and metadata.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[cfg_attr(
+    any(test, feature = "arbitrary"),
+    derive(proptest_derive::Arbitrary, arbitrary::Arbitrary)
+)]
+#[serde(rename_all = "camelCase")]
+pub struct TransactionReceipt<T = ReceiptEnvelope<Log>> {
```
mattsse, line 16: we need to solve this a bit differently
this is basically the ethereum receipt with extra l1 fee fields, but the receipt type is different, so we need a receipt envelope type that can also capture the extra op deposit fields

mattsse (changes requested):
good direction, left a few more suggestions

crates/op-rpc-types/src/op/transaction/request.rs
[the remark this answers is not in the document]
EmperorOrokuSaki: The transaction type has changed to include Optimism's `Deposit` (just noticed it's a typo there and uses TxHash, but still needs to change)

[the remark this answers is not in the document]
EmperorOrokuSaki: https://github.com/EmperorOrokuSaki/op-alloy/blob/fafe2aa0a3e8b044bb309c1684a57c953125ae7f/crates/op-rpc-types/src/op/transaction/request.rs#L59

crates/op-rpc-types/src/op/transaction/receipt.rs
EmperorOrokuSaki: True, I wanted this PR to be just for the RPC types and then make a separate PR for consensus. I'll just add a placeholder for now to not forget this.

crates/op-consensus/src/receipt/envelope.rs
```diff
@@ -0,0 +1,196 @@
...
+
+impl<T> ReceiptEnvelope<T> {
+    /// Return the [`TxType`] of the inner receipt.
+    pub const fn tx_type(&self) -> TxType {
+        match self {
+            Self::Legacy(_) => TxType::Legacy,
+            Self::Eip2930(_) => TxType::Eip2930,
+            Self::Eip1559(_) => TxType::Eip1559,
+            Self::Eip4844(_) => TxType::Eip4844,
+        }
```
gakonst, line 53: isnt this missing the deposit tx? all of these methods should deref to the inner enum variant i think

crates/op-rpc-types/src/op/block.rs
```diff
@@ -0,0 +1,144 @@
...
+        skip_serializing_if = "BlockTransactions::is_uncle"
+    )]
+    pub transactions: BlockTransactions,
+    /// Integer the size of this block in bytes.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub size: Option<U256>,
+    /// Withdrawals in the block.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub withdrawals: Option<Vec<Withdrawal>>,
+}
```
gakonst, line 35: This should just be a re-export of the default json rpc block, no?

crates/op-rpc-types/src/op/filters.rs
```diff
@@ -0,0 +1,69 @@
+use super::transaction::Transaction;
```
gakonst, line 1: reexport

crates/op-rpc-types/src/op/pubsub.rs
```diff
@@ -0,0 +1,37 @@
+//! Optimism types for pub-sub
+
+use crate::op::transaction::Transaction;
+use alloy::rpc::types::eth::{pubsub::PubSubSyncStatus, Log, RichHeader};
+use alloy_primitives::B256;
+use serde::{Deserialize, Serialize, Serializer};
+
+/// Subscription result.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[serde(untagged)]
```
gakonst, line 10: reexport

gakonst (changes requested):
the entire PR needs to re-export RPC types where needed, and extend them everywhere else, the requirement is "there must not be any duplicate types between the 2 crates"

i would start from the lower level crate whether it's consensus or rpc types from your POV and do that first, and you can do the other in separate PR

but def don't copy paste types around

overall good start!

crates/op-rpc-types/src/op/block.rs
EmperorOrokuSaki: The BlockTransactions struct has changed to include the optimism transactions, so I included the block here to use the new BlockTransactions

crates/op-consensus/src/receipt/envelope.rs
EmperorOrokuSaki: This is just a placeholder because consensus should be another pr i think

crates/op-rpc-types/src/op/block.rs
gakonst: we should make that struct take a generic

crates/op-consensus/src/receipt/envelope.rs
gakonst: k let's separate or block the PR on the other work then

EmperorOrokuSaki: Removed the op-consensus crate from this pr completely

crates/op-rpc-types/src/op/transaction/request.rs
mattsse, line 1: I don't think we need this, should be identical to the default TransactionRequest

crates/op-rpc-types/src/op/transaction/receipt.rs
mattsse, line 1: what I want to do here is a type that wraps the regular TransactionReceipt but with the op log specific type, and the additional OP fields as new fields like


```rust
struct TransactionReceipt {
  inner: alloy::TransactionReceipt<>,
l1...
}
```

crates/op-rpc-types/src/op/transaction/mod.rs
```diff
@@ -0,0 +1,252 @@
...
+use self::{request::TransactionRequest, tx_type::TxType};
+
+pub mod receipt;
+pub mod request;
+pub mod tx_type;
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[serde(rename_all = "camelCase")]
+pub struct Transaction {
```
mattsse, line 18: this type should also wrap the alloy transaction type and add the additional fields

crates/op-rpc-types/src/op/pubsub.rs
mattsse, line 1: we need an alloy pr that makes these types generic so we can just reuse them

crates/op-rpc-types/src/op/filters.rs
mattsse, line 1: same here, needs alloy pr to make it generic

crates/op-rpc-types/src/op/call.rs
mattsse, line 1: this can be removed

mattsse (changes requested):

crates/op-rpc-types/src/op/transaction/receipt.rs
EmperorOrokuSaki: Implemented this.

crates/op-rpc-types/src/op/transaction/request.rs
EmperorOrokuSaki: The TxType here has `Deposit` additionally. I can make a PR to alloy to make `TransactionRequest` take a generic for the `TxType` and use that here. Is that a good idea?

EmperorOrokuSaki:
PR for making structs in the main alloy crate take generics: https://github.com/alloy-rs/alloy/pull/573

crates/op-rpc-types/src/op/transaction/request.rs
EmperorOrokuSaki: Resolving: Alloy now takes u8 for tx type

crates/op-consensus/src/receipt/envelope.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// [EIP-1559]: https://eips.ethereum.org/EIPS/eip-1559
+    #[cfg_attr(feature = "serde", serde(rename = "0x2", alias = "0x02"))]
+    Eip1559(ReceiptWithBloom<T>),
+    /// Receipt envelope with type flag 2, containing a [EIP-4844] receipt.
+    ///
+    /// [EIP-4844]: https://eips.ethereum.org/EIPS/eip-4844
+    #[cfg_attr(feature = "serde", serde(rename = "0x3", alias = "0x03"))]
+    Eip4844(ReceiptWithBloom<T>),
+    /// Receipt envelope for Optimism's deposit transactions
+    #[cfg_attr(feature = "serde", serde(rename = "0x7E", alias = "0x7E"))]
```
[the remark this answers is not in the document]
EmperorOrokuSaki: Resolving as the work on consensus will be another pr

EmperorOrokuSaki:
Relevant issue: https://github.com/alloy-rs/op-alloy/issues/5

Cargo.toml
```diff
@@ -12,6 +12,26 @@ homepage = "https://github.com/alloy-rs/op-alloy"
 repository = "https://github.com/alloy-rs/op-alloy"
 exclude = ["benches/", "tests/"]
 
+[workspace.dependencies]
+# Alloy
+alloy-primitives = { version = "0.7.0", default-features = false }
+alloy = { git = "https://github.com/alloy-rs/alloy", rev = "39b8695", features = [
```
zerosnacks, line 18: Make sure to update to 0.7.1 and a recent commit

In the examples repo we have an test that performs an integration test against the latest commit on a nightly basis: https://github.com/alloy-rs/examples/blob/main/.github/workflows/integration.yml which may be useful

crates/op-rpc-types/Cargo.toml
```diff
@@ -10,3 +10,35 @@ homepage.workspace = true
...
 exclude.workspace = true
+
+[dependencies]
+# alloy-rlp = { workspace = true, features = ["arrayvec", "derive"] }
+alloy-primitives = { workspace = true, features = ["rlp", "serde", "std"] }
+# alloy-serde.workspace = true
+# alloy-genesis.workspace = true
+
+# alloy-consensus = { workspace = true, features = ["std", "serde"] }
+# alloy-eips = { workspace = true, features = ["std", "serde"] }
```
zerosnacks, line 21: depending on the situation it may be preferable to access alloy-primitives through the alloy prelude crate with the `rlp` and `serde` flags,`std` is enabled by default

https://github.com/alloy-rs/alloy/blob/main/crates/alloy/src/lib.rs

crates/op-rpc-types/src/op/transaction/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// Ethereum Transaction Types
+    #[serde(flatten)]
+    pub inner: alloy::rpc::types::eth::Transaction,
+    /// The ETH value to mint on L2
+    #[serde(rename = "mint", skip_serializing_if = "Option::is_none")]
+    pub mint: Option<U128>,
```
zerosnacks, line 22: it may be preferable to use Rust's native `u128` type here, that reduces the need for conversions - in line with Alloy's preferences for native types where possible

crates/op-rpc-types/src/op/transaction/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// Hash that uniquely identifies the source of the deposit.
+    #[serde(rename = "sourceHash", skip_serializing_if = "Option::is_none")]
+    pub source_hash: Option<B256>,
+    /// Field indicating whether the transaction is a system transaction, and therefore
+    /// exempt from the L2 gas limit.
+    #[serde(rename = "isSystemTx", skip_serializing_if = "Option::is_none")]
+    pub is_system_tx: Option<bool>,
+    /// Deposit receipt version for deposit transactions post-canyon
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub deposit_receipt_version: Option<U64>,
```
zerosnacks, line 32: Same here

mattsse:
keeping this open because we still need receipt types, but to make incremental progress I've extracted the things we definitely need to #6

can't push here because main branch

PR #4, opened 2024-04-13
feat(op-consensus): `op-alloy-consensus`

clabby:
## Overview

Adds a periphery crate for `alloy-consensus`, `op-alloy-consensus`, that contains OP-specific types for the `TxEnvelope` (to include `TxDeposit`) as well as the `Receipt` + `ReceiptEnvelope`. For all unmodified types, such as the `Header`, consumers should use `alloy-consensus`.

> [!NOTE]
> Blocked by https://github.com/alloy-rs/alloy/pull/529, will update the `alloy` dependency after this has been merged.


mattsse (commented):
great start, 

we should reuse as much as possible from alloy-consensus

clabby:
> great start,
> 
> we should reuse as much as possible from alloy-consensus

Thanks - this should be the minimal amount we need to port, without changing the upstream API of `alloy-consensus`. We can re-use `TxLegacy`, `TxEip2930`, `TxEip1559`, and `TxEip4844`, but would need to expose several of their private functions in `alloy-consensus`:
* `fields_len`
* `decode_signed_fields`
* `encode_with_signature`

crates/op-consensus/src/transaction/optimism.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// The address of the recipient account, or the null (zero-length) address if the deposited
+    /// transaction is a contract creation.
+    #[cfg_attr(feature = "serde", serde(default, skip_serializing_if = "TxKind::is_create"))]
+    pub to: TxKind,
+    /// The ETH value to mint on L2.
+    #[cfg_attr(
+        feature = "serde",
+        serde(
+            default,
+            skip_serializing_if = "Option::is_none",
```
clabby, line 26: Might be handled by `alloy_serde::u128_hex_or_decimal_opt` already

crates/op-consensus/src/receipt/mod.rs
```diff
@@ -0,0 +1,199 @@
+use alloy_primitives::{Bloom, Log};
+
+mod envelope;
+pub use envelope::OpReceiptEnvelope;
+
+mod receipts;
+pub use receipts::{OpReceipt, OpReceiptWithBloom};
+
+/// Receipt is the result of a transaction execution.
+pub trait OpTxReceipt {
```
mattsse, line 10: we already have this in alloy?
minues the deposit functions?

so we could unify the traits and inherit here?

crates/op-consensus/src/receipt/envelope.rs
```diff
@@ -0,0 +1,216 @@
...
+    /// Receipt envelope with type flag 3, containing a [EIP-4844] receipt.
+    ///
+    /// [EIP-4844]: https://eips.ethereum.org/EIPS/eip-4844
+    #[cfg_attr(feature = "serde", serde(rename = "0x3", alias = "0x03"))]
+    Eip4844(OpReceiptWithBloom<T>),
+    /// Receipt envelope with type flag 126, containing a [deposit] receipt.
+    ///
+    /// [deposit]: https://specs.optimism.io/protocol/deposits.html
+    #[cfg_attr(feature = "serde", serde(rename = "0x7E", alias = "0x7E"))]
+    Deposit(OpReceiptWithBloom<T>),
```
mattsse, line 43: hmm, only this variant needs the Op specific receipt type, right?

the others could reuse the the alloy-consensus one?

mattsse (changes requested):
this is great,

only have two questions re receipt stuff

crates/op-consensus/src/receipt/mod.rs
clabby: Yeah good thinking. Done.

crates/op-consensus/src/receipt/envelope.rs
clabby: Yeah, only this variant needs it. Decided to use all `OpReceiptWithBloom`s to avoid type conversion within `as_receipt_with_bloom` + `as_receipt`.

Technically, all of the receipts within OP consensus do use this format, though the extra `deposit_nonce` + `deposit_receipt_version` fields are omitted from the RLP if they don't exist. Could reuse the `ReceiptWithBloom<T>` type for others -- would you rather have the type conversions?

mattsse (changes requested):
hmm, I'd prefer if we could reuse the alloy-consensus receipt envelope type or even the entire variant and then only add an additional deposit variant

adding `pub struct OpReceipt<T = Log> {` doesn't make much sense to me

crates/op-consensus/src/receipt/envelope.rs
```diff
@@ -0,0 +1,216 @@
...
+/// decoded, preserving the presence or absence of the `TransactionType` flag.
+///
+/// Transaction receipt payloads are specified in their respective EIPs.
+///
+/// [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(tag = "type"))]
+#[non_exhaustive]
+pub enum OpReceiptEnvelope<T = Log> {
```
mattsse, line 20: I think we should just reuse the regular alloy consensus types here

crates/op-consensus/src/receipt/receipts.rs
```diff
@@ -0,0 +1,271 @@
...
+use alloy_rlp::{length_of_length, BufMut, Decodable, Encodable};
+
+#[cfg(not(feature = "std"))]
+use alloc::vec::Vec;
+
+/// Receipt containing result of transaction execution.
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+pub struct OpReceipt<T = Log> {
```
mattsse, line 13: this should be `OpDepositReceipt`
which wraps the alloy-consensus receipt and adds extra fields for deposit

mattsse (changes requested):

prestwich:
adds a little work for you, but it's all easy:
https://github.com/alloy-rs/alloy/pull/477

mattsse:
superseded by #8

PR #8, opened 2024-05-20
feat(consensus): op-alloy-consensus

refcell:
**Description**

Maintains commit history with #4 and updates trait types.


crates/op-consensus/src/transaction/optimism.rs
mattsse, line 1: this should be moved to deposit.rs

crates/op-consensus/src/receipt/envelope.rs
```diff
@@ -0,0 +1,216 @@
...
+///
+/// [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(tag = "type"))]
+#[non_exhaustive]
+pub enum OpReceiptEnvelope<T = Log> {
+    /// Receipt envelope with no type flag.
+    #[cfg_attr(feature = "serde", serde(rename = "0x0", alias = "0x00"))]
+    Legacy(OpReceiptWithBloom<T>),
```
mattsse, line 23: hmm, this doesn't feel right to use the Opdepositreceipt for all variants

can we replicate the eth variants and only modify the deposit variant?

mattsse (changes requested):
only one q re receipt variants

crates/op-consensus/src/transaction/envelope.rs
```diff
@@ -0,0 +1,340 @@
...
+
+#[cfg(any(test, feature = "arbitrary"))]
+impl<'a> arbitrary::Arbitrary<'a> for OpTxType {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
+        Ok(match u.int_in_range(0..=3)? {
+            0 => OpTxType::Legacy,
+            1 => OpTxType::Eip2930,
+            2 => OpTxType::Eip1559,
+            3 => OpTxType::Eip4844,
+            127 => OpTxType::Deposit,
```
refcell, line 42: ibid `126`, can we also just match on `DEPOSIT_TX_TYPE_ID`?

crates/op-consensus/src/transaction/envelope.rs
```diff
@@ -0,0 +1,340 @@
...
+impl TryFrom<u8> for OpTxType {
+    type Error = Eip2718Error;
+
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        Ok(match value {
+            0 => Self::Legacy,
+            1 => Self::Eip2930,
+            2 => Self::Eip1559,
+            3 => Self::Eip4844,
+            127 => Self::Deposit,
```
refcell, line 57:
```suggestion
            126 => Self::Deposit,
```

mattsse (approved):

PR #13, opened 2024-06-16
fix(consensus): Re-export Traits

refcell:
**Description**

Since alloy dependencies are imported by github, versioning dependencies can get ugly downstream if a specific revision is imported.

To avoid dependency clashes, re-export necessary traits for downstream use. This allows crates and workspaces like kona to specify external dependency revisions (e.g. importing `alloy_eips` with a specific revision) while also importing the `op-alloy-consensus` on `main` branch along with it's EIP-2718 Encoding and Decoding traits, avoiding a dependency clash with  the `alloy_eips` crate.

Example: https://github.com/ethereum-optimism/kona/pull/267

### DIFF
--- a/crates/op-consensus/src/lib.rs
+++ b/crates/op-consensus/src/lib.rs
@@ -19,6 +19,15 @@
 #[cfg(not(feature = "std"))]
 extern crate alloc;
 
+// Re-export EIP2718 Types and Traits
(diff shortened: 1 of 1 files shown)


gakonst:
versioned releases are out now and should prefer these

PR #17, opened 2024-06-27
feat: extract optimism genesis info

Vid201:
## Motivation

Adds functionalities to extract Optimism genesis data from genesis.

`extract_from` returns `Option` since the extraction can be unsuccessful - negative case should be handled outside.

### DIFF
--- a/crates/rpc-types/src/genesis.rs
+++ b/crates/rpc-types/src/genesis.rs
@@ -1,10 +1,42 @@
 //! OP types for genesis data.
 
-/// Genesis info for Optimism.
+use alloy_serde::OtherFields;
+use serde::de::Error;
+
+/// Container type for all Optimism specific fields in a genesis file.
+#[derive(Default, Debug, Clone, Copy, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct OptimismChainInfo {
+    /// Genesis information
+    pub genesis_info: Option<OptimismGenesisInfo>,
+    /// Base fee information
+    pub base_fee_info: Option<OptimismBaseFeeInfo>,
+}
+
+impl OptimismChainInfo {
+    /// Extracts the Optimism specific fields from a genesis file. These fields are expected to be
+    /// contained in the `genesis.config` under `extra_fields` property.
+    pub fn extract_from(others: &OtherFields) -> Option<Self> {
+        Self::try_from(others).ok()
+    }
+}
+
+impl TryFrom<&OtherFields> for OptimismChainInfo {
+    type Error = serde_json::Error;
+
+    fn try_from(others: &OtherFields) -> Result<Self, Self::Error> {
+        let genesis_info = OptimismGenesisInfo::try_from(others).ok();
+        let base_fee_info = OptimismBaseFeeInfo::try_from(others).ok();
+
+        Ok(OptimismChainInfo { genesis_info, base_fee_info })
+    }
+}
+
+/// The Optimism-specific genesis block specification.
 #[derive(Default, Debug, Clone, Copy, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct OptimismGenesisInfo {
-    /// Bedrock block number
+    /// bedrock block number
     pub bedrock_block: Option<u64>,
     /// regolith hardfork timestamp
     pub regolith_time: Option<u64>,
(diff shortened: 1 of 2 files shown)


mattsse:
yeah, having the sanity test would be great

crates/rpc-types/src/genesis.rs
```diff
@@ -1,5 +1,7 @@
 //! OP types for genesis data.
 
+use alloy_genesis::Genesis;
```
mattsse, line 3: could we import Otherfields from alloy serde instead of depending on genesis?

crates/rpc-types/src/genesis.rs
```diff
@@ -27,3 +29,22 @@ pub struct OptimismBaseFeeInfo {
     /// EIP-1559 denominator after canyon
     pub eip1559_denominator_canyon: Option<u64>,
 }
+
+impl OptimismGenesisInfo {
+    /// Extract the Optimism genesis info from the genesis.
+    pub fn extract_from(genesis: &Genesis) -> Option<Self> {
```
mattsse, line 35: could we add another helper type that has both genesisInfo and feeinfo as fields and then impl fn extract_from(others: &Otherfields)?

and also `Tryfrom<Otherfields>` that simply calls extract?

mattsse (changes requested):

Vid201:
Addressed the comments

crates/rpc-types/src/genesis.rs
```diff
@@ -27,3 +79,172 @@ pub struct OptimismBaseFeeInfo {
...
+}
+
+impl TryFrom<OtherFields> for OptimismBaseFeeInfo {
+    type Error = serde_json::Error;
+
+    fn try_from(others: OtherFields) -> Result<Self, Self::Error> {
+        if let Some(Ok(optimism_base_fee_info)) =
+            others.get_deserialized::<OptimismBaseFeeInfo>("optimism")
+        {
+            Ok(optimism_base_fee_info)
```
mattsse, line 102: this code is duplicated, this should reuse the extract fn

crates/rpc-types/src/genesis.rs
```diff
@@ -27,3 +79,172 @@ pub struct OptimismBaseFeeInfo {
     /// EIP-1559 denominator after canyon
     pub eip1559_denominator_canyon: Option<u64>,
 }
+
+impl OptimismBaseFeeInfo {
+    /// Extract the Optimism base fee info from `OtherFields` object.
```
mattsse, line 84: this should document this will check for an `optimism` object
and is intended to be parsed from the genesis object

crates/rpc-types/src/genesis.rs
```diff
@@ -16,6 +53,21 @@ pub struct OptimismGenesisInfo {
...
+    pub fn extract_from(others: &OtherFields) -> Option<Self> {
+        others.deserialize_as().ok()
+    }
+}
+
+impl TryFrom<OtherFields> for OptimismGenesisInfo {
+    type Error = serde_json::Error;
+
+    fn try_from(others: OtherFields) -> Result<Self, Self::Error> {
+        others.deserialize_as()
```
mattsse, line 67: should delegate to Self::extract_from

crates/rpc-types/src/genesis.rs
```diff
@@ -1,5 +1,42 @@
...
+    type Error = serde_json::Error;
+
+    fn try_from(others: OtherFields) -> Result<Self, Self::Error> {
+        let genesis_info = OptimismGenesisInfo::try_from(others.clone())?;
+        let base_fee_info = OptimismBaseFeeInfo::try_from(others)?;
+
+        Ok(OptimismChainInfo {
+            genesis_info: Some(genesis_info),
+            base_fee_info: Some(base_fee_info),
+        })
```
mattsse, line 36: reuse existing impl

crates/rpc-types/src/genesis.rs
```diff
@@ -1,5 +1,42 @@
 //! OP types for genesis data.
 
+use alloy_serde::OtherFields;
+use serde::de::Error;
+
+/// Info for Optimism chain.
```
mattsse, line 6: this should document that this is a container type for all OP specific fields in a genesis file.

crates/rpc-types/src/genesis.rs
```diff
@@ -1,5 +1,42 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct OptimismChainInfo {
+    /// Genesis info
+    pub genesis_info: Option<OptimismGenesisInfo>,
+    /// Base fee info
+    pub base_fee_info: Option<OptimismBaseFeeInfo>,
+}
+
+impl OptimismChainInfo {
+    /// Extract the Optimism chain info from `OtherFields` object.
```
mattsse, line 17: needs additional context that this is expected to be other, optimism specific fields of `genesis.config`

mattsse (changes requested):
great, a few more nits.

the tryinto are more or less just for convenience but should reuse the existing functions so we don't have the code duplicated.

crates/rpc-types/src/genesis.rs
Vid201: But if we reuse the extract fn we lose the error information (since extract returns option) - thus cannot use `type Error = serde_json::Error`. or should we define custom error enum?

mattsse: I see, then you can flip it the other way around

Vid201:
> great, a few more nits.
> 
> the tryinto are more or less just for convenience but should reuse the existing functions so we don't have the code duplicated.

@mattsse Addressed the comments

mattsse (approved):
nice

PR #35, opened 2024-08-25
feat: Added sync file with types from reth

Redidacove:
Motivation
Issue [reth Issue by matsee ](https://github.com/paradigmxyz/reth/issues/10509)
[Pr comments](https://github.com/paradigmxyz/op-rs/pull/25)
In the pr matsee suggested we should add types here and import there for usage 

 Solution
So as mentioned in issue and pr I moved the types here.
Basically added three types here L1/L2blockref and syncstatus 
## PR Checklist

- [x] Added Types

### DIFF
--- a/crates/rpc-types/src/sync.rs
+++ b/crates/rpc-types/src/sync.rs
@@ -7,7 +7,7 @@ use serde::{Deserialize, Serialize};
 /// The block reference for an L2 block.
 ///
 /// See: <https://github.com/ethereum-optimism/optimism/blob/develop/op-service/eth/id.go#L33>
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct L2BlockRef {
     /// The block hash.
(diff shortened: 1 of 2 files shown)


refcell (approved):

refcell:
Should just be able to run `cargo fmt --all` to fix ci.

crates/rpc-types/src/sync.rs
```diff
@@ -0,0 +1,37 @@
...
+use alloy_rpc_types_eth::BlockId;
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct L2BlockRef {
+    pub hash: Option<B256>,
+    pub number: Option<BlockNumber>,
+    pub parent_hash: Option<B256>,
+    pub timestamp: Option<u64>,
+    pub l1origin: Option<BlockId>,
```
refcell, line 13: Nit, can we make this `l1_origin`? The serde camelCase rename will handle converting it to `l1Origin`

refcell (changes requested):
Quick request

crates/rpc-types/src/sync.rs
mattsse, line 13:
```suggestion
    pub l1_origin: Option<BlockId>,
```

mattsse, line 13: why are these now optional?

crates/rpc-types/src/sync.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub sequence_number: Option<u64>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct L1BlockRef {
+    pub hash: Option<B256>,
+    pub number: Option<BlockNumber>,
+    pub parent_hash: Option<B256>,
+    pub timestamp: Option<u64>,
```
mattsse, line 23: same q

mattsse (changes requested):

PR #37, opened 2024-08-25
feat(rpc-client): Introduce rpc-jsonrpsee Crate

### LINKED ISSUE #32: [Feature] Rpc type for `optimism_outputAtBlock`
### Component

rpc

Add a type for the output of `optimism_outputAtBlock` endpoint.

### Additional context

_No response_

refcell:
**Description**

Introduces an `op-alloy-rpc-jsonrpsee` crate that can replace the non-type logic `reth/optimism/rpc`. Type logic can go in `op-alloy/rpc-types`.

Closes #32


refcell:
> [!NOTE]
>
> Depends on #35

crates/rpc-client/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+
+# rpc
+jsonrpsee.workspace = true
+jsonrpsee-types.workspace = true
+
+[features]
+client = [
+    "jsonrpsee/client",
+    "jsonrpsee/async-client",
+]
```
mattsse, line 27: hmm by default this now generates jsonrpsee server traits, so perhaps `rpc-client` isn't the perfect fit here.

mattsse (changes requested):
perhaps we rename this to op-alloy-jsonrpsee ?

crates/rpc-jsonrpsee/src/traits.rs
```diff
@@ -0,0 +1,31 @@
...
+    #[method(name = "syncStatus")]
+    async fn optimism_sync_status(&self) -> RpcResult<SyncStatus>;
+
+    /// Get the rollup configuration parameters.
+    #[method(name = "rollupConfig")]
+    async fn optimism_rollup_config(&self) -> RpcResult<RollupConfig>;
+
+    /// Get the software version.
+    #[method(name = "version")]
+    async fn optimism_version(&self) -> RpcResult<String>;
```
mattsse, line 30: should we shorten all of these to `op_` and stick with that convention?

refcell: Good callout - I like this a lot more

refcell:
> [!NOTE]
>
> Adds the `OutputResponse` type to `rpc-types` since it's used by the trait. Takes care of #32

HrikB (commented):
I wanna ask if it makes sense to implement a `Provider` here over the `Optimism` network. I was doing smth like this:

```
#[async_trait]
pub trait OpStackProvider<T: Transport + Clone = BoxTransport>: Provider<T, Optimism> {
    async fn output_at_block(&self, block_number: u64) -> TransportResult<OutputAtBlock> {
        let response = self
            .client()
            .request("optimism_outputAtBlock", (format!("0x{:x}", block_number),))
            .await?;

        Ok(response)
    }
}

impl<T: Transport + Clone> OpStackProvider<T> for RootProvider<T, Optimism> {
}
```

mattsse:
yep this makes sense, mind opening an issue for this?
this should go into a new provider crate

mattsse (approved):

PR #42, opened 2024-08-25
feat: add rollup and other config types

mattsse:


refcell (approved):
Very nice - I hope down the road we can re-use the `superchain-primitives` types we have bindings for in the `superchain-regsitry`. 

e.g. https://github.com/ethereum-optimism/superchain-registry/blob/main/bindings/rust-primitives/src/rollup_config.rs

mattsse:
good point, should we even merge this then?

I think not

refcell:
> good point, should we even merge this then?
> 
> I think not

Well... I'm seeing some issues like the `batcher_address` won't serialize nicely in the system config since it's just `camelCase` and won't be renamed to `batcherAddr` like the rpc expects, it seems.

refcell:
We also made the potentially breaking decision in `superchain-primitives` to merge the `ChainConfig` with the `RollupConfig` which I don't believe plays nicely with what the op RPC expects. I think we (myself) ought to go with this for now and investigate a bit more into reusing these types and making sure they're compatible with tests

mattsse:
idk where this is used in rpc, do you mean we should merge?
feel free to make a decision here

refcell:
Cool, I'll merge for now and write up an issue to see if we can re-use `superchain-primitives` types in place of these :)

PR #56, opened 2024-08-28
feat(protocol): Core Protocol Types

refcell:
### Description

Introduce core protocol types to a new `op-alloy` crate called `op-alloy-protocol`.

Nomenclature: `protocol` as opposed to `derive` since these types will be used beyond derivation. They exist as core types used for data transformation between source and destination chains.

This migrates [channel types](https://github.com/ethereum-optimism/kona/blob/main/crates/primitives/src/channel.rs) and [frame types](https://github.com/ethereum-optimism/kona/blob/main/crates/primitives/src/frame.rs) to op-alloy.

> [!NOTE]
>
> This PR will not include batch types.
> Before these can be ported from the `kona-derive::batch` module, batch validity must be trait abstracted in order to avoid pulling in `kona-derive` types.

### DIFF
--- a/crates/protocol/src/block.rs
+++ b/crates/protocol/src/block.rs
@@ -1,6 +1,6 @@
 //! Block Types for Optimism.
 
-use alloy_primitives::B256;
+use alloy_primitives::{U64, B256};
 use superchain_primitives::BlockID;
 
 #[cfg(feature = "serde")]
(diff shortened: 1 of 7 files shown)


crates/protocol/src/block.rs
```diff
@@ -0,0 +1,62 @@
...
+use serde::{Deserialize, Serialize};
+
+/// Block Header Info
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug, Clone, Copy, Eq, Hash, PartialEq, Default)]
+pub struct BlockInfo {
+    /// The block hash
+    pub hash: B256,
+    /// The block number
+    pub number: u64,
```
mattsse, line 16: are these serialized as number or hex?

crates/protocol/src/block.rs
```diff
@@ -0,0 +1,62 @@
...
+            "BlockInfo {{ hash: {}, number: {}, parent_hash: {}, timestamp: {} }}",
+            self.hash, self.number, self.parent_hash, self.timestamp
+        )
+    }
+}
+
+/// L2 Block Header Info
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug, Clone, Copy, Hash, Eq, PartialEq, Default)]
+pub struct L2BlockInfo {
```
mattsse, line 48: camelCase?

mattsse (approved):

Issue #57, opened 2024-08-29
[Bug] Ecotone support missing for `OptimismTransactionReceiptFields`


emhane:
### Component

rpc

### Operating System

None

### Describe the bug

Port new ecotone fields from depreciated alloy type https://github.com/alloy-rs/alloy/commit/ab462f12ac40fd16bd9882c9451d7d62f7c31d37

PR #58, opened 2024-08-29
Add ecotone support to `op_alloy_rpc_types::OptimismTransactionReceiptFileds`

emhane:
## Motivation

Closes https://github.com/alloy-rs/op-alloy/issues/57

## Solution

Port fields from depreciated alloy type https://github.com/alloy-rs/alloy/commit/ab462f12ac40fd16bd9882c9451d7d62f7c31d37

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/rpc-types/src/receipt.rs
+++ b/crates/rpc-types/src/receipt.rs
@@ -85,6 +85,7 @@ pub struct OptimismTransactionReceiptFields {
     /// L1 blob base fee scalar. Applied to blob base fee to compute weighted gas price multiplier.
     ///
     /// Always null prior to the Ecotone hardfork.
+    #[serde(default, skip_serializing_if = "Option::is_none", with = "alloy_serde::quantity::opt")]
     pub l1_blob_base_fee_scalar: Option<u128>,
 }
 
--- a/crates/rpc-types/src/receipt.rs
+++ b/crates/rpc-types/src/receipt.rs
@@ -75,6 +75,7 @@ pub struct OptimismTransactionReceiptFields {
     /// L1 base fee scalar. Applied to base fee to compute weighted gas price multiplier.
     ///
     /// Always null prior to the Ecotone hardfork.
+    #[serde(default, skip_serializing_if = "Option::is_none", with = "alloy_serde::quantity::opt")]
     pub l1_base_fee_scalar: Option<u128>,
     /// L1 blob base fee.
     ///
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm, you're right this should be simplified once we integrate this with the OpTxReceipt, which has the deposit_ fields already.

but for reth we need this as is rn

crates/rpc-types/src/receipt.rs
```diff
@@ -34,29 +34,57 @@ impl ReceiptResponse for OpTransactionReceipt {
...
+    ///
+    /// Always null prior to the Canyon hardfork.
     #[serde(default, skip_serializing_if = "Option::is_none", with = "alloy_serde::quantity::opt")]
-    pub l1_gas_used: Option<u128>,
+    pub deposit_receipt_version: Option<u64>,
+    /* ---------------------------------------- Ecotone ---------------------------------------- */
+    /// L1 base fee scalar. Applied to base fee to compute weighted gas price multiplier.
+    ///
+    /// Always null prior to the Ecotone hardfork.
+    pub l1_base_fee_scalar: Option<u128>,
```
mattsse, line 78: missing attribute
```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none", with = "alloy_serde::quantity::opt")]
    pub l1_base_fee_scalar: Option<u128>,
```

crates/rpc-types/src/receipt.rs
```diff
@@ -34,29 +34,57 @@ impl ReceiptResponse for OpTransactionReceipt {
...
+    /// L1 blob base fee.
+    ///
+    /// Always null prior to the Ecotone hardfork.
+    #[serde(default, skip_serializing_if = "Option::is_none", with = "alloy_serde::quantity::opt")]
+    pub l1_blob_base_fee: Option<u128>,
+    /// L1 blob base fee scalar. Applied to blob base fee to compute weighted gas price multiplier.
+    ///
+    /// Always null prior to the Ecotone hardfork.
+    pub l1_blob_base_fee_scalar: Option<u128>,
```
mattsse, line 87:
```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none", with = "alloy_serde::quantity::opt")]
    pub l1_blob_base_fee_scalar: Option<u128>,
```

mattsse (changes requested):

emhane:
> lgtm, you're right this should be simplified once we integrate this with the OpTxReceipt, which has the deposit_ fields already.
> 
> but for reth we need this as is rn

correct, since we're still using `alloy_rpc_types_eth::AnyTransactionReceipt` in reth RPC

Issue #54, opened 2024-08-28
[Bug] OP fields missing from `op_alloy_rpc_types::OpTransactionReceipt`


emhane:
### Component

consensus, eips, genesis, rpc

### Describe the feature you would like

To add OP fields, either make ~~new variants of `OpReceiptEnvelope`, e.g. `OpReceiptEnvelope::Ecotone`~~ type `OpEcotoneReceipt` for blob fields, or re-define type `OpTransactionReceipt`. Implement decoding of `OpTransactionReceipt` with OP fields. Adding directly to `OpTransactionReceipt` may make sense for some op fields, for others not.

Ref https://github.com/ethereum-optimism/op-geth/blob/f2e69450c6eec9c35d56af91389a1c47737206ca/core/types/receipt.go#L87-L87

### Additional context

Deposit tx envelope has fields for deposit nonce, and so does `OptimismTransactionReceiptFields`, so `OptimismTransactionReceiptFields` cannot be added to `OpTransactionReceipt` as is.

- [ ] Ecotone https://github.com/alloy-rs/op-alloy/pull/53
- [ ] Fjord
list tbc

emhane:
good call @refcell

PR #53, opened 2024-08-28
fix(rpc): impl ecotone for `op_alloy_rpc_types::OpTransactionReceipt`

emhane:
## Motivation

Ref https://github.com/alloy-rs/op-alloy/issues/54

Eip-4844 support missing for `op_alloy_rpc_types::OptimismTransactionReceiptFields` and bad filed naming, causing bugs in reth RPC.

## Solution

- Adds fileds `l1_blob_base_fee` and `l1_blob_base_fee_scalar` to `op_alloy_rpc_types::OptimismTransactionReceiptFields`
- Fixes naming of field `l1_fee_scalar` to `l1_base_fee_scalar`, in coherence with `op-geth` and `reth-evm-optimism`

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

PR #62, opened 2024-08-30
fix(rpc): Add l1 block info to OpTransactionReceipt

emhane:
## Motivation

Closes https://github.com/alloy-rs/op-alloy/issues/54

## Solution

- Moves L1 block info fields out of type `OptimismTransactionReceiptFields` into new type `L1BlockInfo`.
- Adds `L1BlockInfo` to `OpTransactionReceipt`

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/rpc-types/src/receipt.rs
+++ b/crates/rpc-types/src/receipt.rs
@@ -14,6 +14,8 @@ pub struct OpTransactionReceipt {
     /// Regular eth transaction receipt including deposit receipts
     #[serde(flatten)]
     pub inner: alloy_rpc_types_eth::TransactionReceipt<OpReceiptEnvelope<alloy_rpc_types_eth::Log>>,
+    //// L1 block info of the transaction.
+    pub l1_block_info: L1BlockInfo,
 }
 
 impl ReceiptResponse for OpTransactionReceipt {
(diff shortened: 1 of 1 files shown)


refcell (approved):
This makes sense - would like to see a PR down the road that adds doc comments about why these are separate types and how they're used. If you're looking at op-geth, these fields are just part of the receipt and not also a separate OptimismTransactionReceiptFields.

emhane:
> This makes sense - would like to see a PR down the road that adds doc comments about why these are separate types and how they're used. If you're looking at op-geth, these fields are just part of the receipt and not also a separate OptimismTransactionReceiptFields.

the deposit envelope has the deposit nonce and receipt version fields. so adding the `OptimismTransactionReceiptFeilds` type to `OpTransactionReceipt` directly would have caused duplicates for these keys.

I think down the road, this will be replaced with some new envelope types, as https://github.com/alloy-rs/op-alloy/pull/53, to help set the additional op fields. right now it's really quite hard to debug anything related to this since I have to memorise which fields have to be set for which fork, from reading docs on the `OptimismTransactionReceiptFeilds` (now `L1BlockInfo`) type.

the type `OptimismTransactionReceiptFields` will probably be depreciated when we can stop using the `AnyTransactionReceipt` type in reth https://github.com/paradigmxyz/reth/issues/10632, which is possible after next `op-alloy` release to include this pr. when the linked issue is solved, we can just call a builder for `OpTransactionReceipt` directly, no need for intermediary type `OptimismTransactionReceiptFields` to convert to `OtherFields`.

PR #67, opened 2024-09-02
fix(docs): l1 gas used deprecated since Fjord not Ecotone

emhane:
## Motivation

Docs bug: l1 gas used is deprecated since Fjord not Ecotone
https://github.com/ethereum-optimism/op-geth/blob/f2e69450c6eec9c35d56af91389a1c47737206ca/core/types/receipt.go#L90

## Solution

Update docs

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/receipt.rs
+++ b/crates/rpc-types/src/receipt.rs
@@ -107,7 +107,7 @@ pub struct L1BlockInfo {
     pub l1_gas_price: Option<u128>,
     /// L1 gas used.
     ///
-    /// Present from pre-bedrock to Ecotone. Null after Ecotone.
+    /// Present from pre-bedrock to Ecotone. Null after Fjord.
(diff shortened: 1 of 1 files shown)


crates/rpc-types/src/receipt.rs
```diff
@@ -107,7 +107,7 @@ pub struct L1BlockInfo {
     pub l1_gas_price: Option<u128>,
     /// L1 gas used.
     ///
-    /// Present from pre-bedrock to Ecotone. Null after Ecotone.
+    /// Present from pre-bedrock to Ecotone. Null after Fjord.
```
mattsse, line 110: can we mirror op-geth's docs 1:1 here

> Present from pre-bedrock, deprecated as of Fjord

mattsse (changes requested):

mattsse (approved):

PR #73, opened 2024-09-05
fix(protocol): Native u64

refcell:
### Description

Uses the native `u64` type - this prevents a lot of type conversions in kona.

See https://github.com/anton-rs/kona/pull/495

### DIFF
--- a/crates/protocol/src/block.rs
+++ b/crates/protocol/src/block.rs
@@ -1,6 +1,6 @@
 //! Block Types for Optimism.
 
-use alloy_primitives::{B256, U64};
+use alloy_primitives::B256;
 use superchain_primitives::BlockID;
 
 #[cfg(feature = "serde")]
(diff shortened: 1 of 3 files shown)


crates/protocol/src/block.rs
```diff
@@ -53,12 +55,39 @@ pub struct L2BlockInfo {
     /// The L1 origin [BlockID]
     pub l1_origin: BlockID,
     /// The sequence number of the L2 block
-    pub seq_num: U64,
+    #[cfg_attr(feature = "serde", serde(deserialize_with = "deserialize_u64_hex"))]
+    pub seq_num: u64,
```
Rjected, line 59: I think all of these can be replaced with
```
    #[serde(with = "alloy_serde::quantity")]
```

Rjected (changes requested):
one comment on alloy serde

crates/protocol/src/block.rs
```diff
@@ -53,12 +55,39 @@ pub struct L2BlockInfo {
...
+}
+
+#[cfg(feature = "serde")]
+fn deserialize_u64_hex<'de, D>(deserializer: D) -> Result<u64, D::Error>
```
mattsse, line 63: yeah this is just alloy_serde::quantity

mattsse (changes requested):

crates/protocol/src/block.rs
refcell: Amazing thank you

Rjected (approved):
nice

PR #74, opened 2024-09-05
feat(protocol): Deposit Tx Utility

refcell:
### Description

Adds a simple utility method to tell if a tx is a deposit transaction.

### DIFF
--- a/crates/protocol/src/utils.rs
+++ b/crates/protocol/src/utils.rs
@@ -5,7 +5,7 @@ pub fn is_deposit<B>(value: &B) -> bool
 where
     B: AsRef<[u8]>,
 {
-    !value.as_ref().is_empty() && value.as_ref()[0] == 0x7E
+    value.as_ref().first() == Some(&0x7E)
 }
 
 #[cfg(test)]
--- a/crates/protocol/src/utils.rs
+++ b/crates/protocol/src/utils.rs
@@ -1,0 +1,21 @@
+//! Utility methods used by protocol types.
+
+/// Returns if the given `value` is a deposit transaction.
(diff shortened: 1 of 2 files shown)


crates/protocol/src/utils.rs
```diff
@@ -0,0 +1,29 @@
+//! Utility methods used by protocol types.
+
+/// Returns if the given `value` is a deposit transaction.
+pub fn is_deposit<B>(value: &B) -> bool
+where
+    B: AsRef<[u8]>,
+{
+    !value.as_ref().is_empty() && value.as_ref()[0] == 0x7E
```
mattsse, line 8:
```suggestion
    value.as_ref().first() == Some(&0x7E)
```

crates/protocol/src/utils.rs
```diff
@@ -0,0 +1,29 @@
+//! Utility methods used by protocol types.
+
+/// Returns if the given `value` is a deposit transaction.
+pub fn is_deposit<B>(value: &B) -> bool
```
mattsse, line 4: perhaps rename this to `starts_with_2781_deposit` or smth?

mattsse (changes requested):
also fyi:

https://github.com/alloy-rs/alloy/blob/fb7b09ed92c5f938a36803e9fda688c8a99bcbe2/crates/eips/src/eip2718.rs#L64-L69

crates/protocol/src/utils.rs
refcell: Ah nice, will do

mattsse (approved):

PR #80, opened 2024-09-07
fix(protocol): EIP 2718 Deposit Tx Utility

refcell:
### Description

Fixes the deposit util to be for EIP 2718.

Renames:
`starts_with_2781_deposit` to
`starts_with_2718_deposit`


mattsse (approved):
haha

PR #81, opened 2024-09-08
Feat(op-alloy-protocol): add deposit module 

DoTheBestToGetTheBest:
related https://github.com/anton-rs/kona/issues/509

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,85 +1,40 @@
-[workspace]
-members = ["crates/*"]
-resolver = "2"
+[package]
+name = "op-alloy-protocol"
+description = "Optimism protocol-specific types"
 
-[workspace.package]
-version = "0.2.8"
-edition = "2021"
-rust-version = "1.79"
-authors = ["Alloy Contributors"]
-license = "MIT OR Apache-2.0"
-homepage = "https://github.com/alloy-rs/op-alloy"
-repository = "https://github.com/alloy-rs/op-alloy"
-exclude = ["benches/", "tests/"]
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+authors.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+exclude.workspace = true
 
-[workspace.lints]
-rust.unused_must_use = "deny"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -71,7 +79,7 @@ thiserror = "1.0"
 proptest = "1.4"
 proptest-derive = "0.4"
 tokio = "1"
-
+anyhow = { version = "1.0.86", default-features = false }
```
refcell, line 82: Let's avoid using `anyhow`

crates/protocol/Cargo.toml
```diff
@@ -14,25 +14,28 @@ exclude.workspace = true
...
 
+# Op-Alloy
+op-alloy-consensus.workspace = true
+
 # `serde` feature dependencies
 serde = { workspace = true, optional = true }
 alloy-serde = { workspace = true, optional = true }
 
+
+anyhow.workspace = true
```
refcell, line 34: Remove anyhow

crates/protocol/src/deposits.rs
```diff
@@ -0,0 +1,688 @@
...
+    /// An invalid deposit version.
+    InvalidVersion(B256),
+    /// Unexpected opaque data length
+    UnexpectedOpaqueDataLen(usize),
+    /// Failed to decode the deposit mint value.
+    MintDecode(Bytes),
+    /// Failed to decode the deposit gas value.
+    GasDecode(Bytes),
+    /// A custom error wrapping [anyhow::Error].
+    Custom(anyhow::Error),
```
refcell, line 57: I think we should use a generic here that implements the [`core::error::Error`](https://doc.rust-lang.org/core/error/trait.Error.html) trait.

e.g.
```rust
pub enum DepositError<T: impl core::error::Error> {
	...
	/// A custom [core::error::Error].
	Custom(T)
}
```

refcell (changes requested):
Request to remove `anyhow` as a dependency.

DoTheBestToGetTheBest:
ty for the fast review; made all changes you requested

refcell (approved):

mattsse:
this did not pass CI

PR #84, opened 2024-09-08
feat(rpc-types): Replace u8 with Connectedness Enum

DoTheBestToGetTheBest:
Replaced u8 types with the Connectedness enum 

ref of the implementation https://github.com/libp2p/go-libp2p/blob/d55bed5f78fad923353b92c7d1db64c758bcbf9c/core/network/network.go#L49

related to https://github.com/ethereum-optimism/optimism/blob/40750a58e7a4a6f06370d18dfe6c6eab309012d9/op-node/p2p/rpc_api.go#L25

### DIFF
--- a/crates/rpc-types/src/net.rs
+++ b/crates/rpc-types/src/net.rs
@@ -2,7 +2,7 @@
 //! Network RPC types
 
 use alloy_primitives::ChainId;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::{collections::HashMap, net::IpAddr};
 
 // https://github.com/ethereum-optimism/optimism/blob/8dd17a7b114a7c25505cd2e15ce4e3d0f7e3f7c1/op-node/p2p/store/iface.go#L13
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
this makes sense, but this should ensure that this is serialized/deserialized as u8, which default serde derive doesn't do afaik.

so I'd also like to see a test for this

crates/rpc-types/src/net.rs
```diff
@@ -100,3 +100,54 @@ pub struct PeerStats {
     pub banned: u32,
     pub known: u32,
 }
+
+/// Represents the connectivity state of a peer in a network, indicating the reachability and
+/// interaction status of a node with its peers.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Copy)]
+#[serde(rename_all = "camelCase")]
```
refcell, line 107: We'll want to add the u8 representation. Also should probably mark `NotConnected` as the `#[default]`

```rust
#[repr(u8)]
```

crates/rpc-types/src/net.rs
```diff
@@ -100,3 +100,160 @@ pub struct PeerStats {
...
+
+impl<'de> Deserialize<'de> for Connectedness {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let value = u8::deserialize(deserializer)?;
+        Ok(Connectedness::from(value))
+    }
+}
```
refcell, line 173: We should be able to remove these and place `#[derive(Serialize, Deserialize)]` above `Connectedness`.

DoTheBestToGetTheBest: > We should be able to remove these and place `#[derive(Serialize, Deserialize)]` above `Connectedness`.

Yes but the tests are failling if i derive the Serialize and Deserialize

refcell: Let me reproduce and check it out

refcell (approved):

PR #87, opened 2024-09-09
chore: cleanup depositerror

mattsse:
ref #81

this removes the generic, doesn't seem reasonable.
and fixes ci


mattsse:
@refcell in case you need a custom error consider adding it again

PR #88, opened 2024-09-09
fix(protocol): Functional Batch Transaction

refcell:
### Description

Fixes up the `BatchTransaction` to use a more functional pattern.


crates/protocol/src/batch_tx.rs
```diff
@@ -29,10 +29,34 @@ impl BatchTransaction {
...
+        let frame = Frame { id: [0xFF; 16], number: 0xEE, data: vec![0xDD; 50], is_last: true };
+        let batch = BatchTransaction { frames: vec![frame.clone(); 5], size: 5 * frame.size() };
+
+        let bytes: Bytes = (&batch).into();
+        let bytes =
+            [crate::DERIVATION_VERSION_0].iter().chain(bytes.iter()).copied().collect::<Vec<_>>();
+        let frames = Frame::parse_frames(&bytes).unwrap();
+        assert_eq!(frames.len(), 5);
+        (0..5).for_each(|i| {
+            assert_eq!(frames[i], frame);
```
emhane, line 59: guessing that `Frame` doesn't impl `Eq`?

emhane (approved):

crates/protocol/src/batch_tx.rs
```diff
@@ -29,10 +29,34 @@ impl BatchTransaction {
 
 impl From<&BatchTransaction> for Bytes {
```
mattsse, line 30: I personally very much dislike these kinds of conversions, this should just be a function instead.

this way you get proper intellisense, docs and don't have to do `(&v).into()`

mattsse (approved):
optional suggestion

PR #89, opened 2024-09-09
feat(protocol): payload module

DoTheBestToGetTheBest:
Since we dont use anyhow anymore we made new type for error handling

### DIFF
--- a/crates/protocol/src/block.rs
+++ b/crates/protocol/src/block.rs
@@ -6,6 +6,10 @@ use alloy_primitives::B256;
 use op_alloy_consensus::{OpBlock, OpTxEnvelope, OpTxType};
 use op_alloy_genesis::ChainGenesis;
 
+use alloy_consensus::Header;
+use alloy_eips::eip4895::Withdrawal;
+use alloy_rlp::{RlpDecodable, RlpEncodable};
+use op_alloy_consensus::OpTxEnvelope;
 /// Block Header Info
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[derive(Debug, Clone, Copy, Eq, Hash, PartialEq, Default)]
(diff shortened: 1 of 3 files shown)


crates/protocol/src/block.rs
```diff
@@ -62,7 +66,23 @@ impl L2BlockInfo {
...
+pub struct OpBlock {
+    /// Block header.
+    pub header: Header,
+    /// Transactions in this block.
+    pub body: Vec<OpTxEnvelope>,
+    /// Ommers/uncles header.
+    pub ommers: Vec<Header>,
+    /// Block withdrawals.
+    pub withdrawals: Option<Vec<Withdrawal>>,
+}
```
refcell, line 85: We want to use https://github.com/alloy-rs/alloy/blob/main/crates/rpc-types-eth/src/block.rs#L20 instead of duplicating the block definition here. We can then alias `Block<OpTxEnvelope>` as `OpBlock`. The issue right now is `alloy-rpc-types-eth` doesn't support `no_std` _and_ requires serde by default. I have a pr open for this here: https://github.com/alloy-rs/alloy/pull/1252

crates/protocol/src/payload.rs
```diff
@@ -0,0 +1,314 @@
...
+            out += tx.len() as u64 + PAYLOAD_TX_MEM_OVERHEAD;
+        }
+        out
+    }
+}
+
+/// The execution payload.
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct L2ExecutionPayload {
```
refcell, line 94: We should be able to avoid duplicating this type and use an alloy execution payload type as well.

refcell: Here's the `ExecutionPayload` type: https://github.com/alloy-rs/alloy/blob/4492c25f1a666881d18367dc47d0040fc05b17ae/crates/rpc-types-engine/src/payload.rs#L594

refcell: I also have a pr open in upstream alloy to support `no_std` in `alloy-rpc-types-engine`: https://github.com/alloy-rs/alloy/pull/1233

crates/protocol/src/block.rs
DoTheBestToGetTheBest: > We want to use https://github.com/alloy-rs/alloy/blob/main/crates/rpc-types-eth/src/block.rs#L20 instead of duplicating the block definition here. We can then alias `Block<OpTxEnvelope>` as `OpBlock`. The issue right now is `alloy-rpc-types-eth` doesn't support `no_std` _and_ requires serde by default. I have a pr open for this here: [alloy-rs/alloy#1252](https://github.com/alloy-rs/alloy/pull/1252)

thank you for your fast reply, i'll wait your pr get merged to use Block type then!

crates/protocol/src/payload.rs
DoTheBestToGetTheBest: Thanks for sharing this,

 Should we wait for your changes to get merged upstream before we integrate this into the crate?

refcell:
This payload now exists in https://github.com/alloy-rs/op-alloy/blob/main/crates/rpc-types-engine/src/payload_v3.rs

PR #90, opened 2024-09-11
replace u8 direction field with Direction type

DoTheBestToGetTheBest:
according to https://github.com/ethereum-optimism/optimism/blob/40750a58e7a4a6f06370d18dfe6c6eab309012d9/op-node/p2p/rpc_api.go#L26C24-L26C38

implementation reference https://github.com/libp2p/go-libp2p/blob/d55bed5f78fad923353b92c7d1db64c758bcbf9c/core/network/network.go#L26


crates/rpc-types/src/net.rs
```diff
@@ -149,7 +149,42 @@ impl From<u8> for Connectedness {
...
+#[repr(u8)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Default)]
+pub enum Direction {
+    /// DirUnknown is the default direction when the direction is not specified.
+    #[default]
+    DirUnknown = 0,
+    /// DirInbound is for when the remote peer initiated the connection.
+    DirInbound = 1,
+    /// DirOutbound is for when the local peer initiated the connection.
+    DirOutbound = 2,
```
mattsse, line 162: lets drop the Dir prefix

crates/rpc-types/src/net.rs
```diff
@@ -149,7 +149,42 @@ impl From<u8> for Connectedness {
         }
     }
 }
+/// Direction represents the direction of a connection.
+#[repr(u8)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Default)]
```
mattsse, line 154: does this serde as number?

please add a test

crates/rpc-types/src/net.rs
```diff
@@ -149,7 +149,42 @@ impl From<u8> for Connectedness {
...
+impl From<u8> for Direction {
+    fn from(num: u8) -> Self {
+        match num {
+            0 => Direction::DirInbound,
+            1 => Direction::DirOutbound,
+            2 => Direction::DirUnknown,
+            _ => Direction::DirInbound,
+        }
+    }
+}
```
mattsse, line 173: don't think we can do this, mapping unsupported numbers to inbound doesn't feel right, please convert to tryfrom

mattsse (changes requested):

DoTheBestToGetTheBest:
done  @mattsse thanks

mattsse (approved):

PR #91, opened 2024-09-12
Make `l1_origin` in `L2BlockRef` a struct instead of an enum

HrikB:
Currently, `l1_origin` in the `L2BlockRef` struct is an enum. But from the link embedded above it, it is supposed to be a struct. 

See: https://github.com/ethereum-optimism/optimism/blob/develop/op-service/eth/id.go#L10-L13

### DIFF
--- a/crates/rpc-types/src/sync.rs
+++ b/crates/rpc-types/src/sync.rs
@@ -1,17 +1,8 @@
 //! Op types related to sync.
 
+use alloy_eips::BlockNumHash;
 use alloy_primitives::{BlockNumber, B256};
 use serde::{Deserialize, Serialize};
-
-
-/// See: https://github.com/ethereum-optimism/optimism/blob/develop/op-service/eth/id.go#L10-L13
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
-pub struct BlockID {
-    /// The block hash.
-    pub hash: B256,
-    /// The block number.
-    pub number: BlockNumber,
-}
 
 /// The block reference for an L2 block.
 ///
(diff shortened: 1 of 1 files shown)


crates/rpc-types/src/sync.rs
```diff
@@ -1,9 +1,18 @@
...
 
+
+/// See: https://github.com/ethereum-optimism/optimism/blob/develop/op-service/eth/id.go#L10-L13
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct BlockID {
+    /// The block hash.
+    pub hash: B256,
+    /// The block number.
+    pub number: BlockNumber,
+}
```
mattsse, line 14: this can be alloy's BlockNumHash type

mattsse (changes requested):

crates/rpc-types/src/sync.rs
HrikB: It says `BlockNumHash` doesn't implement `Deserialize`

mattsse: added https://github.com/alloy-rs/alloy/pull/1277

HrikB: Should I set the `rev` to that commit or wait for a release?

mattsse: can put this on hold after you changed it to blockNumHash until the next release

HrikB: So just leave it for now?

mattsse: > can put this on hold after you changed it to blockNumHash until the next release

refcell (approved):

mattsse:
@refcell this can be closed?

refcell:
> @refcell this can be closed?

I think this is still valid actually

mattsse (approved):

PR #94, opened 2024-09-12
fix: Remove Block ID

refcell:
### Description

Re-uses [`alloy_eips::BlockNumHash`](https://docs.rs/alloy-eips/latest/alloy_eips/eip1898/type.BlockNumHash.html) instead of a custom `BlockID` type in `superchain-primitives`.

This follows #91

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,9 @@ option_if_let_else = "warn"
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
+[patch.crates-io]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -27,6 +27,9 @@ option_if_let_else = "warn"
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
+[patch.crates-io]
+alloy-eips = { git = "https://github.com/alloy-rs/alloy", branch = "main" }
+
```
refcell, line 32: TODO: remove once `BlockNumHash` has serde support.

mattsse (approved):

PR #110, opened 2024-09-17
feat(engine): Payload Conversion Utilities

refcell:
### Description

Implements default payload conversion utils used in `kona`.

### DIFF
--- a/crates/rpc-types-engine/src/lib.rs
+++ b/crates/rpc-types-engine/src/lib.rs
@@ -40,13 +40,13 @@ mod errors;
 pub use errors::{ToL2BlockRefError, ToSystemConfigError};
 
 /// Converts a payload into its most inner [ExecutionPayloadV1].
-pub trait IntoInnerPayload {
+pub trait AsInnerPayload {
     /// Converts the payload into its most inner [ExecutionPayloadV1].
-    fn inner_payload(&self) -> &crate::ExecutionPayloadV1;
+    fn as_v1_payload(&self) -> alloc::borrow::Cow<'_, ExecutionPayloadV1>;
 }
 
-/// OptimismPayload trait defines conversion methods for Optimism-specific payloads.
-pub trait OptimismPayload {
+/// Defines conversion utility methods for Optimism-specific payloads.
+pub trait OptimismPayloadUtils {
     /// Converts the payload into an [L2BlockInfo].
     fn to_l2_block_ref(
         &self,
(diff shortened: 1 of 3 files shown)


clabby (approved):

crates/rpc-types-engine/src/errors.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// Failed to decode the first transaction into an [op_alloy_consensus::OpTxEnvelope].
+    TxEnvelopeDecodeError(Eip2718Error),
+    /// The first payload transaction is not a deposit transaction.
+    FirstTxNonDeposit(u8),
+    /// Failed to decode the [op_alloy_protocol::L1BlockInfoTx] from the deposit transaction.
+    BlockInfoDecodeError(DecodeError),
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for ToL2BlockRefError {}
```
mattsse, line 27: this can be core now

crates/rpc-types-engine/src/errors.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// The first payload transaction is not a deposit transaction.
+    FirstTxNonDeposit(u8),
+    /// Failed to decode the [op_alloy_protocol::L1BlockInfoTx] from the deposit transaction.
+    BlockInfoDecodeError(DecodeError),
+    /// Missing system config in the genesis block.
+    MissingSystemConfig,
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for ToSystemConfigError {}
```
mattsse, line 75: same

crates/rpc-types-engine/src/errors.rs
```diff
@@ -0,0 +1,103 @@
...
+
+impl core::fmt::Display for ToSystemConfigError {
```
mattsse, line 77: optional, you can use derive_more display for this

crates/rpc-types-engine/src/lib.rs
```diff
@@ -18,5 +18,154 @@
...
+pub use payload_v3::OptimismExecutionPayloadEnvelopeV3;
+
+mod payload_v4;
+pub use payload_v4::OptimismExecutionPayloadEnvelopeV4;
+
+mod errors;
+pub use errors::{ToL2BlockRefError, ToSystemConfigError};
+
+/// Converts a payload into its most inner [ExecutionPayloadV1].
+pub trait IntoInnerPayload {
```
mattsse, line 43: this naming is a bit misleading, because this only borrows

crates/rpc-types-engine/src/lib.rs
```diff
@@ -18,5 +18,154 @@
...
+    /// Converts the payload into its most inner [ExecutionPayloadV1].
+    fn inner_payload(&self) -> &crate::ExecutionPayloadV1;
```
mattsse, line 45: can we keep versioning in function name if this returns V1

we can also return `Cow` here so that it's also possible to create this type if this can't be returned as `&`

mattsse (changes requested):
some suggestions

crates/rpc-types-engine/src/lib.rs
refcell: Agreed, fixing.

mattsse (approved):

PR #112, opened 2024-09-18
feat(engine): Deprecate RollupConfig Argument

refcell:
### Description

Deprecates the need to pass a reference to the full `RollupConfig` object, using the more lightweight `ChainGenesis` object.

### DIFF
--- a/crates/rpc-types-engine/src/lib.rs
+++ b/crates/rpc-types-engine/src/lib.rs
@@ -48,24 +48,6 @@ pub trait AsInnerPayload {
 /// Defines conversion utility methods for Optimism-specific payloads.
 pub trait OptimismPayloadUtils {
     /// Converts the payload into an [L2BlockInfo].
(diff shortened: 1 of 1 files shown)


crates/rpc-types-engine/src/lib.rs
```diff
@@ -48,55 +48,75 @@ pub trait AsInnerPayload {
...
 
     /// Converts the payload into a [SystemConfig].
+    #[deprecated(since = "0.2.12", note = "Use `to_sys_config` instead")]
     fn to_system_config(
         &self,
         rollup_config: &RollupConfig,
-    ) -> Result<SystemConfig, ToSystemConfigError>;
+    ) -> Result<SystemConfig, ToSystemConfigError> {
+        self.to_sys_config(&rollup_config.genesis)
+    }
```
mattsse, line 66: I'm fine with just removing those

mattsse (approved):

PR #114, opened 2024-09-18
fix(engine): Missing Error Source

refcell:
### Description

Fixes `op-alloy-rpc-types-engine` conversion errors to provide their source for 2718 decoding errors and block info decoding errors.

### DIFF
--- a/crates/rpc-types-engine/src/errors.rs
+++ b/crates/rpc-types-engine/src/errors.rs
@@ -29,7 +29,15 @@ pub enum ToL2BlockRefError {
     BlockInfoDecodeError(DecodeError),
 }
 
-impl core::error::Error for ToL2BlockRefError {}
+impl core::error::Error for ToL2BlockRefError {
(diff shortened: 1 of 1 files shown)


emhane (approved):
nice code comment. would you open an alloy issue to bump the rust version if one doesn't exists?

crates/rpc-types-engine/src/errors.rs
```diff
@@ -58,4 +81,27 @@ pub enum ToSystemConfigError {
...
+        }
+    }
+}
+
+#[cfg(not(feature = "std"))]
+impl core::error::Error for ToSystemConfigError {
+    fn source(&self) -> Option<&(dyn core::error::Error + 'static)> {
+        match self {
+            Self::BlockInfoDecodeError(err) => Some(err),
+            _ => None,
```
emhane, line 104: just checking, this is purposely missing a variant?

PR #117, opened 2024-09-21
feat(provider): OP engine api trait ext + superchain signal type

merklefruit:
## Motivation

Adds a `provider` crate (`op-alloy-provider`) with some extension API traits for the OP engine API.
Additionally, adds the `SuperchainSignal` type along with the canonical decoding impl for the V0 type.

Ref: https://github.com/paradigmxyz/op-rs/issues/11

### DIFF
--- a/crates/rpc-types-engine/Cargo.toml
+++ b/crates/rpc-types-engine/Cargo.toml
@@ -24,6 +24,7 @@ derive_more = { workspace = true, features = ["display"] }
 # serde
 serde = { workspace = true, optional = true }
 alloy-serde = { workspace = true, optional = true }
+cfg-if = { version = "1.0" }
 
 [dev-dependencies]
 serde_json.workspace = true
--- a/crates/rpc-types-engine/Cargo.toml
+++ b/crates/rpc-types-engine/Cargo.toml
@@ -20,7 +20,7 @@ op-alloy-genesis.workspace = true
 op-alloy-consensus.workspace = true
 
 derive_more = { workspace = true, features = ["display"] }
-thiserror.workspace = true
+thiserror = { workspace = true, optional = true }
 
 # serde
 serde = { workspace = true, optional = true }
@@ -32,6 +32,7 @@ serde_json.workspace = true
 [features]
 default = ["std", "serde"]
 std = [
+  "dep:thiserror",
   "alloy-primitives/std",
   "alloy-rpc-types-engine/std",
   "op-alloy-protocol/std",
--- a/crates/rpc-types-engine/Cargo.toml
+++ b/crates/rpc-types-engine/Cargo.toml
@@ -20,7 +20,7 @@ op-alloy-genesis.workspace = true
 op-alloy-consensus.workspace = true
 
 derive_more = { workspace = true, features = ["display"] }
-cfg-if.workspace = true
+thiserror.workspace = true
 
 # serde
 serde = { workspace = true, optional = true }
--- a/crates/rpc-types-engine/Cargo.toml
+++ b/crates/rpc-types-engine/Cargo.toml
@@ -20,7 +20,6 @@ op-alloy-genesis.workspace = true
 op-alloy-consensus.workspace = true
 
 derive_more = { workspace = true, features = ["display"] }
-thiserror = { workspace = true, optional = true }
 
 # serde
 serde = { workspace = true, optional = true }
@@ -32,7 +31,6 @@ serde_json.workspace = true
 [features]
 default = ["std", "serde"]
 std = [
-  "dep:thiserror",
   "alloy-primitives/std",
   "alloy-rpc-types-engine/std",
   "op-alloy-protocol/std",
--- a/crates/rpc-types-engine/Cargo.toml
+++ b/crates/rpc-types-engine/Cargo.toml
@@ -20,11 +20,11 @@ op-alloy-genesis.workspace = true
 op-alloy-consensus.workspace = true
 
 derive_more = { workspace = true, features = ["display"] }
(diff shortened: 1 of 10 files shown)


crates/rpc-types-engine/src/superchain.rs
```diff
@@ -0,0 +1,146 @@
...
+/// protocol version. The version must be encoded as 32 bytes of DATA in JSON RPC usage.
+///
+/// See also: <https://specs.optimism.io/protocol/superchain-upgrades.html#protocol-version>
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum ProtocolVersion {
+    /// Version-type 0.
+    V0(ProtocolVersionFormatV0),
+}
+
+cfg_if! {
```
mattsse, line 45: I don't think this is required

crates/rpc-types-engine/src/superchain.rs
```diff
@@ -0,0 +1,146 @@
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct ProtocolVersionFormatV0 {
+    pub build: B64,
+    pub major: u32,
+    pub minor: u32,
+    pub patch: u32,
+    pub pre_release: u32,
+}
+
+impl From<ProtocolVersionFormatV0> for [u8; 31] {
```
mattsse, line 101: let's impl this as function, from conversions like this are just obfuscating things.

crates/rpc-types-engine/src/superchain.rs
```diff
@@ -0,0 +1,146 @@
...
+        // <minor> ::= <big-endian uint32>
+        // <patch> ::= <big-endian uint32>
+        // <pre-release> ::= <big-endian uint32>
+
+        Ok(Self {
+            build: B64::from_slice(&value[7..15]),
+            major: u32::from_be_bytes(value[15..19].try_into()?),
+            minor: u32::from_be_bytes(value[19..23].try_into()?),
+            patch: u32::from_be_bytes(value[23..27].try_into()?),
+            pre_release: u32::from_be_bytes(value[27..31].try_into()?),
```
mattsse, line 143: this will panic if called with a slice < len32

needs an additional len check

mattsse (changes requested):

crates/rpc-types-engine/Cargo.toml
```diff
@@ -20,7 +20,7 @@ op-alloy-genesis.workspace = true
 op-alloy-consensus.workspace = true
 
 derive_more = { workspace = true, features = ["display"] }
-cfg-if.workspace = true
+thiserror.workspace = true
```
mattsse, line 23: if you want no-std then this won't work.

I'd suggest impl via derive_more

crates/rpc-types-engine/src/superchain.rs
```diff
@@ -0,0 +1,180 @@
...
+pub enum ProtocolVersionError {
+    #[error("Unsupported version: {0}")]
+    UnsupportedVersion(u8),
+    #[error("Invalid version format length. Got {0}, expected 31")]
+    InvalidLength(usize),
+    #[error("Invalid version format encoding")]
+    FromSlice(#[from] TryFromSliceError),
+}
+
+impl From<ProtocolVersion> for B256 {
```
mattsse, line 55: please convert these to functions, these from conversions are horrible to work with.

crates/rpc-types-engine/src/superchain.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// ```text
+    /// <reserved><build><major><minor><patch><pre-release>
+    /// <reserved> ::= <7 zeroed bytes>
+    /// <build> ::= <8 bytes>
+    /// <major> ::= <big-endian uint32>
+    /// <minor> ::= <big-endian uint32>
+    /// <patch> ::= <big-endian uint32>
+    /// <pre-release> ::= <big-endian uint32>
+    /// ```
+    pub fn into_slice(self) -> [u8; 31] {
```
mattsse, line 145: this is not a slice.

perhaps just encode?

mattsse (changes requested):
last nits

crates/rpc-types-engine/src/superchain.rs
merklefruit: Got it, I thought the `[u8; 31]` was the issue

crates/rpc-types-engine/src/superchain.rs
merklefruit: renamed to `encode` and `decode` respectively

crates/rpc-types-engine/Cargo.toml
merklefruit: Done with derive_more

crates/rpc-types-engine/src/superchain.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// Identifies additional functionality in backwards compatible manner
+    pub minor: u32,
+    /// Identifies backward-compatible bug-fixes
+    pub patch: u32,
+    /// Identifies unstable versions that may not satisfy the above
+    pub pre_release: u32,
+}
+
+#[cfg(feature = "std")]
+impl std::fmt::Display for ProtocolVersionFormatV0 {
```
mattsse, line 180: this can be core::fmt::Display

mattsse (approved):
final nit hehe

merklefruit:
@mattsse done and added decoding tests from op-geth test vectors.
WASM target is failing for something related to "getrandom" lib - not sure how to address this

PR #118, opened 2024-09-24
feat(protocol): Holocene Support

refcell:
### Description

Updates the [`L1BlockInfoTx`](https://github.com/alloy-rs/op-alloy/blob/main/crates/protocol/src/block_info.rs#L41) in `op-alloy-protocol` to include EIP 1559 denominator and elasticity parameters. This extends the `L1BlockInfoTx` with a new `Holocene` variant.

See the [Holocene L1 block attributes specs](https://specs.optimism.io/protocol/holocene/l1-attributes.html#l1-block-attributes) for the transaction byte representation.

### Additional Context

See https://github.com/anton-rs/kona/issues/564


mattsse (approved):
lgtm

PR #134, opened 2024-10-01
Adding fee computation functions to l1BlockInfoTx

cody-wang-cb:
## Motivation

We want to reuse `op-alloy`'s [block info type ](https://github.com/alloy-rs/op-alloy/blob/main/crates/protocol/src/block_info.rs#L47 ) in `revm`, but it's missing some [fee computation functions](https://github.com/bluealloy/revm/blob/ca5b2fe15e0b9920ede13f1c67918b93bfdc12c2/crates/optimism/src/l1block.rs#L128-L249). Thus this PR aims to add these functionalities.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/protocol/src/block_info.rs
```diff
@@ -566,6 +614,70 @@ impl L1BlockInfoEcotone {
...
+        l1_fee_scaled
+            .saturating_mul(rollup_data_gas_cost)
+            .wrapping_div(U256::from(1_000_000 * NON_ZERO_BYTE_COST))
+    }
+
+    /// Calculate the gas cost of a transaction based on L1 block data posted on L2, pre-Ecotone.
+    pub fn calculate_tx_l1_cost_bedrock(&self, input: &[u8]) -> U256 {
+        let rollup_data_gas_cost = self.data_gas(input);
+        rollup_data_gas_cost
+            //.saturating_add(self.l1_fee_overhead)
```
cody-wang-cb, line 664: Not very sure what's the best way to deal with `l1_fee_overhead` for post-`ecotone` blocks, is it worth to add the attribute to the struct?

crates/protocol/src/utils.rs
```diff
@@ -8,22 +7,134 @@ where
...
+    fn test_flz_compress_len_no_repeats() {
+        let mut input = Vec::new();
+        let mut len = 0;
+
+        for i in 0..256 {
+            input.push(i as u8);
+            let prev_len = len;
+            len = flz_compress_len(&input);
+            assert!(len > prev_len);
+        }
```
cody-wang-cb, line 138: It seems challenging to port this test over https://github.com/bluealloy/revm/blob/ca5b2fe15e0b9920ede13f1c67918b93bfdc12c2/crates/optimism/src/fast_lz.rs#L154
since it reads the EVM state. 

Wondering if there's any suggestions for this

crates/protocol/src/block_info.rs
```diff
@@ -500,6 +516,38 @@ impl L1BlockInfoBedrock {
...
+    pub fn data_gas(&self, input: &[u8]) -> U256 {
+        let mut rollup_data_gas_cost = U256::from(input.iter().fold(0, |acc, byte| {
+            acc + if *byte == 0x00 { ZERO_BYTE_COST } else { NON_ZERO_BYTE_COST }
+        }));
+
+        // Prior to regolith, an extra 68 non zero bytes were included in the rollup data costs.
+        rollup_data_gas_cost += U256::from(NON_ZERO_BYTE_COST).mul(U256::from(68));
+
+        rollup_data_gas_cost
+    }
```
clabby, line 536: This is incorrect - the `data_gas` function in `L1BlockInfoBedrock` needs a separate variant for pre-Regolith to add the signature cost overhead (`NON_ZERO_BYTE_COST * 68`). After regolith, this value was removed from the cost.

crates/protocol/src/block_info.rs
```diff
@@ -500,6 +516,38 @@ impl L1BlockInfoBedrock {
...
+
+    /// Calculate the gas cost of a transaction based on L1 block data posted on L2, pre-Ecotone.
+    pub fn calculate_tx_l1_cost(&self, input: &[u8]) -> U256 {
+        if input.is_empty() || input.first() == Some(&0x7F) {
+            return U256::ZERO;
+        }
```
clabby, line 542: This can be moved into the high-level impl in `L1BlockInfoTx`.

crates/protocol/src/block_info.rs
```diff
@@ -566,6 +614,70 @@ impl L1BlockInfoEcotone {
...
+        let rollup_data_gas_cost = self.data_gas(input);
+        let l1_fee_scaled = self.calculate_l1_fee_scaled_ecotone();
+
+        l1_fee_scaled
+            .saturating_mul(rollup_data_gas_cost)
+            .wrapping_div(U256::from(1_000_000 * NON_ZERO_BYTE_COST))
+    }
+
+    /// Calculate the gas cost of a transaction based on L1 block data posted on L2, pre-Ecotone.
+    pub fn calculate_tx_l1_cost_bedrock(&self, input: &[u8]) -> U256 {
```
clabby, line 661: I'd like it if we didn't duplicate this. Can we just move these `calculate_tx_l1_cost_<fork>` fns out of the variants?

crates/protocol/src/block_info.rs
```diff
@@ -642,6 +754,56 @@ impl L1BlockInfoHolocene {
             eip_1559_elasticity,
         })
     }
+
+    /// Calculate the data gas for posting the transaction on L1. Calldata costs 16 gas per byte
+    /// after compression.
+    pub fn data_gas(&self, input: &[u8]) -> U256 {
```
clabby, line 760: This also doesn't seem like it'll work - The `Ecotone` variant needs to support the fjord tx cost function, which is before this variant will be active.

clabby (changes requested):

crates/protocol/src/utils.rs
clabby: We can add `revm` as a `dev-dependency` to use it for this test.

crates/protocol/src/block_info.rs
cody-wang-cb: Is line 533 not doing this? Or maybe I'm not understanding

crates/protocol/src/block_info.rs
clabby: This function should only ever be used on a `L1BlockInfoBedrock` variant. When the type is decoded from calldata or storage, the variant is decided. The fee functions on each variant need to support the forks that span their activity, i.e. `L1BlockInfoEcotone` needs to support ecotone + fjord changes, and `L1BlockInfoBedrock` needs to support bedrock + regolith changes.

At the time that we decode them in `revm`, we'll decide the same way we do today, with checks like the [empty scalars check](https://github.com/bluealloy/revm/blob/ca5b2fe15e0b9920ede13f1c67918b93bfdc12c2/crates/optimism/src/l1block.rs#L108-L115). When decoding from calldata, we can switch on the function selector.

crates/protocol/src/block_info.rs
cody-wang-cb: Yeah duplicated the `ecotone` method below because of this. Let me look into how to de-dup these methods.

crates/protocol/src/block_info.rs
clabby: It is performing the pre-regolith behavior, but there's nowhere to enable the regolith fork's behavior, which uses this variant of the L1 info transaction. The specs for the change are [here](https://specs.optimism.io/protocol/exec-engine.html#pre-ecotone), see "Before Regolith fork..." and "With Regolith fork..."

crates/protocol/src/fee.rs
```diff
@@ -0,0 +1,483 @@
...
+/// Calculate the data gas for posting the transaction on L1.
+///
+/// In regolith, calldata costs 16 gas per non-zero byte and 4 gas per zero byte
+fn data_gas_regolith(input: &[u8]) -> U256 {
+    let rollup_data_gas_cost = U256::from(input.iter().fold(0, |acc, byte| {
+        acc + if *byte == 0x00 { ZERO_BYTE_COST } else { NON_ZERO_BYTE_COST }
+    }));
+
+    rollup_data_gas_cost
+}
```
clabby, line 37:
```suggestion
/// Calculate the data gas for posting the transaction on L1.
///
/// In bedrock, calldata costs 16 gas per non-zero byte and 4 gas per zero byte, with
/// an extra 68 non-zero bytes were included in the rollup data costs to account for the empty signature.
fn data_gas_bedrock(input: &[u8]) -> U256 {
    data_gas_regolith(input) + U256::from(NON_ZERO_BYTE_COST).mul(U256::from(68))
}

/// Calculate the data gas for posting the transaction on L1.
///
/// In regolith, calldata costs 16 gas per non-zero byte and 4 gas per zero byte
fn data_gas_regolith(input: &[u8]) -> U256 {
    let rollup_data_gas_cost = U256::from(input.iter().fold(0, |acc, byte| {
        acc + if *byte == 0x00 { ZERO_BYTE_COST } else { NON_ZERO_BYTE_COST }
    }));

    rollup_data_gas_cost
}
```

crates/protocol/src/fee.rs
```diff
@@ -0,0 +1,483 @@
...
+/// `calldataGas*(l1BaseFee*16*l1BaseFeeScalar + l1BlobBaseFee*l1BlobBaseFeeScalar)/16e6`
+pub fn calculate_tx_l1_cost_ecotone(
+    input: &[u8],
+    l1_fee_overhead: U256,
+    l1_fee_scalar: U256,
+    base_fee: U256,
+    base_fee_scalar: U256,
+    blob_base_fee: U256,
+    blob_base_fee_scalar: U256,
+    empty_scalars: bool,
```
clabby, line 118: I don't think we need to take this into this function; the consumer can decide to use the Bedrock fee function at a higher level.

clabby:
> One thing I’m still not sure about is the edge case around the first block for `ecotone` needs to use `calculate_tx_l1_cost_bedrock` calculation, which requires `l1_fee_overhead`. So in order to implement `calculate_tx_l1_cost` for `L1BlockInfoTxEcotone` it seems like it must add this attribute?

I think they should not. The consumer can decide to handle this case; i.e. in revm, we'll check if there's empty scalars like we do today, and then choose which of these functions we use as a result.

cody-wang-cb:
@clabby just addressed the comments and all the checks passed now, PTAL again

refcell (approved):
This is looking good to me pending @clabby's review

clabby (approved):

emhane (approved):
lgtm, would be nice if you can open an issue to use these to integrate/replace equivalent logic in `reth_optimism_evm::l1`

PR #138, opened 2024-10-01
feat(rpc-types-engine): EIP-1559 parameters in `OptimismPayloadAttributes`

clabby:
## Overview

Adds the `eip1559Params` field to the `OptimismPayloadAttributes, per https://github.com/ethereum-optimism/specs/pull/404

### DIFF
--- a/crates/rpc-types-engine/src/attributes.rs
+++ b/crates/rpc-types-engine/src/attributes.rs
@@ -1,7 +1,7 @@
 //! Optimism-specific payload attributes.
 
 use alloc::vec::Vec;
-use alloy_primitives::{Bytes, FixedBytes};
+use alloy_primitives::{Bytes, B64};
 use alloy_rpc_types_engine::PayloadAttributes;
 use op_alloy_protocol::L2BlockInfo;
 
@@ -30,7 +30,7 @@ pub struct OptimismPayloadAttributes {
     ///
     /// Prior to Holocene activation, this field should always be [None].
     #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
-    pub eip_1559_params: Option<FixedBytes<8>>,
+    pub eip_1559_params: Option<B64>,
 }
 
 /// Optimism Payload Attributes with parent block reference.
--- a/crates/rpc-types-engine/src/attributes.rs
+++ b/crates/rpc-types-engine/src/attributes.rs
@@ -70,3 +70,54 @@ impl OptimismAttributesWithParent {
         self.is_last_in_span
     }
 }
+
+#[cfg(all(test, feature = "serde"))]
+mod test {
+    use super::*;
+    use alloy_primitives::{b64, Address, B256};
+    use alloy_rpc_types_engine::PayloadAttributes;
+
+    #[test]
+    fn test_serde_roundtrip_attributes_pre_holocene() {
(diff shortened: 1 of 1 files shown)


crates/rpc-types-engine/src/attributes.rs
```diff
@@ -26,6 +26,11 @@ pub struct OptimismPayloadAttributes {
         serde(skip_serializing_if = "Option::is_none", with = "alloy_serde::quantity::opt")
     )]
     pub gas_limit: Option<u64>,
+    /// If set, this sets the EIP-1559 parameters for the block.
+    ///
+    /// Prior to Holocene activation, this field should always be [None].
+    #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
+    pub eip_1559_params: Option<FixedBytes<8>>,
```
mattsse, line 33: this is u8 and can use alloy_serde::quantity::opt

mattsse (changes requested):

crates/rpc-types-engine/src/attributes.rs
mattsse: or is it?
maybe not, if this is a hash type then this should be `B8`

clabby: It's not? It's not - it's a fixed bytes type of `8` bytes in len.

extension is spec'd as:
```
PayloadAttributesV3: {
    timestamp: QUANTITY
    random: DATA (32 bytes)
    suggestedFeeRecipient: DATA (20 bytes)
    withdrawals: array of WithdrawalV1
    parentBeaconBlockRoot: DATA (32 bytes)
    transactions: array of DATA
    noTxPool: bool
    gasLimit: QUANTITY or null
    eip1559Params: DATA (8 bytes)
}
```

we're packing 2x 32 bit values (denominator, elasticity) within it.

clabby: changed to `B64`.

crates/rpc-types-engine/src/attributes.rs
```diff
@@ -26,6 +26,11 @@ pub struct OptimismPayloadAttributes {
         serde(skip_serializing_if = "Option::is_none", with = "alloy_serde::quantity::opt")
     )]
     pub gas_limit: Option<u64>,
+    /// If set, this sets the EIP-1559 parameters for the block.
+    ///
+    /// Prior to Holocene activation, this field should always be [None].
+    #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
+    pub eip_1559_params: Option<B64>,
```
mattsse, line 33: could we add a serde roundtrip test where some fields are set to none?

suddenly no longer 100% sure whether we don't actually need serde(default) here

crates/rpc-types-engine/src/attributes.rs
mattsse: gotcha

crates/rpc-types-engine/src/attributes.rs
refcell: +1

refcell: Though I don't think we do need a `serde(default)` since this follows the above optional fields.

refcell (approved):

mattsse (approved):

PR #139, opened 2024-10-01
feat(genesis): `SystemConfig` holocene updates

clabby:
## Overview

Adds the new EIP-1559 parameters to the `SystemConfig` in preparation for the Holocene hardfork.

New config update type is specified [here](https://specs.optimism.io/protocol/holocene/configurability.html#configupdate).

### DIFF
--- a/crates/genesis/src/system.rs
+++ b/crates/genesis/src/system.rs
@@ -512,7 +512,7 @@ pub enum EIP1559UpdateError {
     LengthDecodingError,
     /// The data length is invalid.
     InvalidDataLength(u64),
-    /// Failed to decode the gas limit argument from the eip 1559 update log.
+    /// Failed to decode the eip1559 params argument from the eip 1559 update log.
     EIP1559DecodingError,
 }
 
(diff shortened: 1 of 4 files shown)


crates/genesis/src/system.rs
```diff
@@ -453,6 +498,54 @@ impl core::fmt::Display for GasLimitUpdateError {
...
+    InvalidDataLen(usize),
+    /// Failed to decode the data pointer argument from the eip 1559 update log.
+    PointerDecodingError,
+    /// The data pointer is invalid.
+    InvalidDataPointer(u64),
+    /// Failed to decode the data length argument from the eip 1559 update log.
+    LengthDecodingError,
+    /// The data length is invalid.
+    InvalidDataLength(u64),
+    /// Failed to decode the gas limit argument from the eip 1559 update log.
```
refcell, line 515: I think this is bad copy pasta - should be
```suggestion
    /// Failed to decode the eip1559 params argument from the eip 1559 update log.
```

refcell (approved):
One nit, otherwise GTG

PR #144, opened 2024-10-04
feat(genesis): Arbitrary Support

refcell:
### Description

In [`kona`](https://github.com/anton-rs/kona), we'd like to test against arbitrary inputs for various points in the derivation pipeline. implementing arbitrary across op-alloy input types enables this.

This PR implements [`arbitrary::Arbitrary`](https://docs.rs/arbitrary/latest/arbitrary/trait.Arbitrary.html) across all `op-alloy-genesis` crate types.

### DIFF
--- a/crates/genesis/src/rollup.rs
+++ b/crates/genesis/src/rollup.rs
@@ -140,6 +140,44 @@ pub struct RollupConfig {
     pub da_challenge_address: Option<Address>,
 }
 
+#[cfg(any(test, feature = "arbitrary"))]
+impl<'a> arbitrary::Arbitrary<'a> for RollupConfig {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+        Ok(Self {
+            genesis: ChainGenesis::arbitrary(u)?,
+            block_time: u.arbitrary()?,
+            max_sequencer_drift: u.arbitrary()?,
+            seq_window_size: u.arbitrary()?,
+            channel_timeout: u.arbitrary()?,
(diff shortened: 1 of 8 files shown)


crates/genesis/src/rollup.rs
```diff
@@ -140,6 +140,44 @@ pub struct RollupConfig {
...
+        Ok(Self {
+            genesis: ChainGenesis::arbitrary(u)?,
+            block_time: u.arbitrary()?,
+            max_sequencer_drift: u.arbitrary()?,
+            seq_window_size: u.arbitrary()?,
+            channel_timeout: u.arbitrary()?,
+            granite_channel_timeout: u.arbitrary()?,
+            l1_chain_id: u.arbitrary()?,
+            l2_chain_id: u.arbitrary()?,
+            base_fee_params: BaseFeeParams {
```
mattsse, line 155: could we derive this if this were arbitrary?

mattsse (approved):

crates/genesis/src/rollup.rs
refcell: Yes

mattsse: we can easily add

refcell: I'm happy to take that on if you'd like

mattsse: yes please

perhaps this should use the regular op or eth params always

refcell: Oh good point... will put up an issue for this so as to not block.

PR #147, opened 2024-10-04
fix(genesis): BaseFeeParams Arbitrary Bounds

refcell:
### Description

Updates the arbitrary `BaseFeeParams` to bound to types. 

### Context

See [discussion here](https://github.com/alloy-rs/op-alloy/pull/144#discussion_r1787856319)


clabby (approved):

mattsse (approved):
right, this makes more sense

PR #149, opened 2024-10-05
feat: op network execution payload envelope decoding

merklefruit:
## Motivation

Ports the necessary types for decoding a snappy+ssz encoded `ExecutionPayloadEnvelope`  from the OP CL network.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation


crates/rpc-types-engine/src/payload.rs
```diff
@@ -0,0 +1,175 @@
+use alloy_primitives::{keccak256, Signature, B256};
+use alloy_rpc_types_engine::{
+    ExecutionPayload, ExecutionPayloadV1, ExecutionPayloadV2, ExecutionPayloadV3,
+};
+use derive_more::derive::{Display, From};
+use ssz::Decode;
+
+/// Optimism execution payload envelope in network format.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct OptimismNetworkExecutionPayloadEnvelope {
```
merklefruit, line 10: This name is a little confusing, what is a more idiomatic name? 
essentially this is the CL network representation of unsafe blocks as they get gossiped by the sequencer.

crates/rpc-types-engine/src/payload.rs
```diff
@@ -0,0 +1,175 @@
...
+
+impl PayloadHash {
+    /// The expected message that should be signed by the unsafe block signer.
+    pub fn signature_message(&self, chain_id: u64) -> B256 {
+        let domain = B256::ZERO.as_slice();
+        let chain_id = B256::left_padding_from(&chain_id.to_be_bytes()[..]);
+        let payload_hash = self.0.as_slice();
+        keccak256([domain, chain_id.as_slice(), payload_hash].concat())
+    }
+}
```
refcell, line 128: I really thought we had this somewhere but I've looked and can't seem to find it. I'm ok adding it and refactoring it out later if we find it down the road.

crates/rpc-types-engine/src/payload.rs
refcell: Tough. Maybe just `OptimismNetworkPayloadEnvelope`? No real need for the "execution" provision

crates/rpc-types-engine/Cargo.toml
```diff
@@ -15,18 +15,26 @@ exclude.workspace = true
 workspace = true
 
 [dependencies]
+snap = "1.1.1"
```
refcell, line 18: [snap](https://github.com/BurntSushi/rust-snappy) is licensed under the `BSD-3 Clause`. We need to include this license to publish a crate with this dep I believe.

refcell: This should be as simple as adding [the license](https://github.com/BurntSushi/rust-snappy/blob/master/COPYING) as a new file in the repo root and referencing it in the module level doc comment in `payload.rs`.

crates/rpc-types-engine/src/payload.rs
merklefruit: Ah, I took it from here: https://github.com/paradigmxyz/op-rs/blob/bdd68f73cce39643c285292be18a56f85f9a4c46/crates/net/src/types/payload.rs#L10

crates/rpc-types-engine/src/payload.rs
```diff
@@ -0,0 +1,175 @@
...
+    #[from(alloy_primitives::SignatureError)]
+    InvalidSignature,
+    /// The SSZ encoding is broken.
+    #[display("Broken SSZ encoding")]
+    #[from(ssz::DecodeError)]
+    BrokenSszEncoding,
+    /// The payload envelope is of invalid length.
+    #[display("Invalid length")]
+    InvalidLength,
+}
```
mattsse, line 39: nit: move this below the impl block

crates/rpc-types-engine/src/payload.rs
```diff
@@ -0,0 +1,175 @@
...
+
+        if decompressed.len() < 66 {
+            return Err(PayloadEnvelopeError::InvalidLength);
+        }
+
+        let sig_data = &decompressed[..65];
+        let block_data = &decompressed[65..];
+
+        let signature = Signature::try_from(sig_data)?;
+        let hash = PayloadHash::from(block_data);
```
mattsse, line 77: this is fine because from_ssz_bytes expects exact length

crates/rpc-types-engine/src/payload.rs
mattsse: at some point we need to do a naming audit, because we currently use both `Optimism` and `Op` prefixes

refcell: Noticed that as well... I should be able to take a pass at this early this week!

refcell: https://github.com/alloy-rs/op-alloy/issues/150

crates/rpc-types-engine/src/envelope.rs
```diff
@@ -0,0 +1,183 @@
...
+    BrokenSszEncoding,
+    /// The payload envelope is of invalid length.
+    #[display("Invalid length")]
+    InvalidLength,
+}
+
+/// Represents the Keccak256 hash of the block
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+pub struct PayloadHash(B256);
```
mattsse, line 119: this should perhaps be pub, or at least we need a new fn I think

mattsse (approved):

refcell (approved):

crates/rpc-types-engine/src/envelope.rs
```diff
@@ -0,0 +1,194 @@
...
+use alloy_primitives::{keccak256, Signature, B256};
+use alloy_rpc_types_engine::ExecutionPayload;
+use derive_more::derive::{Display, From};
+
+/// Optimism execution payload envelope in network format.
+///
+/// This struct is used to represent payloads that are sent over the Optimism
+/// CL p2p network in a snappy-compressed format.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct OptimismNetworkPayloadEnvelope {
```
emhane, line 15:
```suggestion
pub struct OpNetworkPayloadEnvelope {
```
recent new naming convention

emhane (changes requested):

crates/rpc-types-engine/src/envelope.rs
```diff
@@ -0,0 +1,194 @@
...
+
+impl PayloadHash {
+    /// The expected message that should be signed by the unsafe block signer.
+    pub fn signature_message(&self, chain_id: u64) -> B256 {
+        let domain = B256::ZERO.as_slice();
+        let chain_id = B256::left_padding_from(&chain_id.to_be_bytes()[..]);
+        let payload_hash = self.0.as_slice();
+        keccak256([domain, chain_id.as_slice(), payload_hash].concat())
+    }
+}
```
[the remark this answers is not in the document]
merklefruit: bump here @mattsse

refcell (approved):

crates/rpc-types-engine/src/envelope.rs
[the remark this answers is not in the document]
mattsse: returning the hash is is perfectly reasonable, the PayloadId is something else

mattsse (approved):

refcell (approved):

PR #121, opened 2024-09-24
Replace `UnsignedTx` and `TxEnvelope` with op types in `Network` impl

emhane:
## Motivation

Bug: deposit tx not covered by alloy transaction envelope and unsigned transaction types in `Network` impl for `Optimism`

## Solution

Replaces alloy types for `Network::TxEnvelope` and `Network::UnsignedTx` with OP types `OpTxEnvelope` and `OpTypedTransaction` respectively.

blocked by https://github.com/alloy-rs/op-alloy/issues/27

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

Issue #27, opened 2024-07-28
[Feature] Impl `TransactionBuilder::complete_type` for `Optimism`


emhane:
### Component

rpc

### Describe the feature you would like

Add type `OpTransactionRequest` that wraps a `TransactionRequest`. Implement method `OpTransactionRequest::complete_deposit`, for using in impl of `TransactionBuilder::complete_type` for `Optimism`.

emhane:
@refcell would you be keen on this?

refcell:
> @refcell would you be keen on this?

Certainly, I'll take a look later today!

PR #151, opened 2024-10-05
fix: remove 4844 transaction type

klkvr:
## Motivation

EIP-4844 transactions are not supported on OP stack networks https://specs.optimism.io/protocol/exec-engine.html#ecotone-disable-blob-transactions

## Solution

Removes 4844 variants from transaction and receipt consensus types.

We still have those on RPC types, should we introduce op-specific RPC types? it would also unblock https://github.com/alloy-rs/op-alloy/pull/121

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/receipt/envelope.rs
+++ b/crates/consensus/src/receipt/envelope.rs
@@ -32,11 +32,6 @@ pub enum OpReceiptEnvelope<T = Log> {
     /// [EIP-1559]: https://eips.ethereum.org/EIPS/eip-1559
     #[cfg_attr(feature = "serde", serde(rename = "0x2", alias = "0x02"))]
     Eip1559(ReceiptWithBloom<T>),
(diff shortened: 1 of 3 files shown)


emhane (approved):
nice!

emhane:
> We still have those on RPC types, should we introduce op-specific RPC types? it would also unblock #121
> 
yepp valid, pls open an issue and link in the blocked issue

klkvr:
> > We still have those on RPC types, should we introduce op-specific RPC types? it would also unblock #121
> 
> yepp valid, pls open an issue and link in the blocked issue

looks like we already track it in https://github.com/alloy-rs/op-alloy/issues/27

don't think think we should duplicate the types completely, can just keep extra fields as `None` and don't implement 7702 and 4844 builders for `OpTransactionRequest`

PR #152, opened 2024-10-07
feat(`consensus`): impl `SignableTx` for `TxDeposit`

yash-atreya:
## Motivation

Ref: https://github.com/foundry-rs/foundry/pull/9047 

## Solution

impl `SignableTransaction<alloy_primitives::Signature>` for `TxDeposit`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/deposit.rs
+++ b/crates/consensus/src/transaction/deposit.rs
@@ -1,4 +1,6 @@
 use super::OpTxType;
+#[cfg(not(feature = "std"))]
+use alloc::vec::Vec;
 use alloy_consensus::{SignableTransaction, Signed, Transaction};
 use alloy_eips::eip2930::AccessList;
 use alloy_primitives::{keccak256, Address, Bytes, ChainId, TxKind, B256, U256};
--- a/crates/consensus/src/transaction/deposit.rs
+++ b/crates/consensus/src/transaction/deposit.rs
@@ -1,10 +1,9 @@
-use super::{OpTxType, DEPOSIT_TX_TYPE_ID};
+use super::OpTxType;
 use alloy_consensus::{SignableTransaction, Signed, Transaction};
 use alloy_eips::eip2930::AccessList;
 use alloy_primitives::{keccak256, Address, Bytes, ChainId, TxKind, B256, U256};
 use alloy_rlp::{
-    length_of_length, Buf, BufMut, Decodable, Encodable, Error as DecodeError, Header,
-    EMPTY_STRING_CODE,
+    Buf, BufMut, Decodable, Encodable, Error as DecodeError, Header, EMPTY_STRING_CODE,
 };
 use core::mem;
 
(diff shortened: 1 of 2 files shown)


crates/consensus/src/transaction/deposit.rs
```diff
@@ -241,6 +243,28 @@ impl Decodable for TxDeposit {
     }
 }
 
+impl SignableTransaction<alloy_primitives::Signature> for TxDeposit {
```
emhane, line 246:
```suggestion
impl SignableTransaction<Signature> for TxDeposit {
```
nitpick, bring this into scope in module import prelude

crates/consensus/src/transaction/deposit.rs
```diff
@@ -241,6 +243,28 @@ impl Decodable for TxDeposit {
...
+
+    fn payload_len_for_signature(&self) -> usize {
+        self.encoded_len(false)
+    }
+
+    fn encode_for_signing(&self, out: &mut dyn alloy_rlp::BufMut) {
+        self.encode_inner(out, false);
+    }
+
+    fn into_signed(self, signature: alloy_primitives::Signature) -> alloy_consensus::Signed<Self> {
```
emhane, line 257:
```suggestion
    fn into_signed(self, signature: Signature) -> alloy_consensus::Signed<Self> {
```
nitpick

emhane (approved):

crates/consensus/src/transaction/deposit.rs
```diff
@@ -241,6 +243,28 @@ impl Decodable for TxDeposit {
...
+    fn encode_for_signing(&self, out: &mut dyn alloy_rlp::BufMut) {
+        self.encode_inner(out, false);
+    }
+
+    fn into_signed(self, signature: Signature) -> alloy_consensus::Signed<Self> {
+        let payload_length = self.encoded_len(false) + signature.rlp_vrs_len();
+        let mut out = Vec::with_capacity(payload_length);
+        self.encode_inner(&mut out, false);
+        signature.write_rlp_vrs(&mut out);
+        let hash = keccak256(out);
```
klkvr, line 262: this will produce invalid hash. deposit transaction rlp encoding does not include a signature

klkvr (commented):
deposit transactions are not "signable" and do not carry a signature, why do we need this?

klkvr:
I see, we migrated this from anvil

I don't think we need this for it because signature is always getting dropped when converting to envelope https://github.com/foundry-rs/foundry/blob/47f1ecb9c6f7e251c5bf2452c1f327d5508481a9/crates/anvil/src/eth/sign.rs#L143

imo even if it's needed in anvil we shouldn't introduce trait implementations which produce off-spec encodings

yash-atreya:
Yeah @klkvr, my mistake. 

Revert PR: https://github.com/alloy-rs/op-alloy/pull/153

PR #157, opened 2024-10-07
fix: Genesis l1 l2 field types

mattsse:
same as https://github.com/alloy-rs/op-alloy/pull/156


refcell (approved):

mattsse:
@refcell updated the test because this one looked incorrect:

https://github.com/ethereum-optimism/optimism/blob/77c91d09eaa44d2c53bec60eb89c5c55737bc325/op-node/rollup/types_test.go#L32-L34

PR #159, opened 2024-10-08
feat(consensus): Transaction for OpTxEnvelope

refcell:
### Description

Implements the [`alloy_consensus::Transaction`](https://docs.rs/alloy-consensus/latest/alloy_consensus/transaction/trait.Transaction.html) trait for the `OpTxEnvelope` type.

This exposes a gas limit accessor that should be used by `kona-executor` instead of [its custom method](https://github.com/anton-rs/kona/blob/main/crates/executor/src/util.rs#L61-L71).


clabby (approved):

refcell:
Will wait for mattsse to give approval :)

mattsse (changes requested):
looks like we're actually missing this impl:

https://github.com/alloy-rs/alloy/blob/de2a13340303db6dde1296f68df57078a5b1fdd8/crates/consensus/src/transaction/envelope.rs#L369-L369

we should do that instead

refcell:
> alloy-rs/alloy@`de2a133`/crates/consensus/src/transaction/envelope.rs#L369-L369

Oh nice! Will do.

mattsse (approved):

PR #165, opened 2024-10-08
chore(consensus): Upstream Receipt Constructor

refcell:
### Description

Small PR to upstream a [receipt constructor from kona](https://github.com/anton-rs/kona/blob/main/crates/executor/src/util.rs#L11).

### DIFF
--- a/crates/consensus/src/receipt/envelope.rs
+++ b/crates/consensus/src/receipt/envelope.rs
@@ -39,6 +39,57 @@ pub enum OpReceiptEnvelope<T = Log> {
     Deposit(OpDepositReceiptWithBloom<T>),
 }
 
+/// Compute the logs bloom filter for the given logs.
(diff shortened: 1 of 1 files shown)


crates/consensus/src/receipt/envelope.rs
```diff
@@ -44,6 +45,57 @@ pub enum OpReceiptEnvelope<T = Log> {
...
+pub(crate) fn logs_bloom<'a>(logs: impl IntoIterator<Item = &'a Log>) -> Bloom {
+    let mut bloom = Bloom::ZERO;
+    for log in logs {
+        bloom.m3_2048(log.address.as_slice());
+        for topic in log.topics() {
+            bloom.m3_2048(topic.as_slice());
+        }
+    }
+    bloom
+}
```
mattsse, line 58: we should actually just move this to alloy-primitives

fine here for now

mattsse:
fyi this fn should now be on latest alloy-primitives

mattsse (approved):

PR #167, opened 2024-10-09
feat: add 7702

mattsse:
this might appear premature, but we need this already


klkvr (approved):
do we need 4844 back too? removed in #151

clabby:
wdyt about feature flagging this until it's live? Not opposed though.

rkrasiuk (approved):
checks out

rkrasiuk:
@klkvr nope, 4844 should not exist for op

shekhirin (approved):

mattsse:
> wdyt about feature flagging this until it's live? Not opposed though.

can discuss, but urgently need this, we have a better workaround for this, but needs more time

mattsse:
@clabby I definitely want to remove these again, but this unblocks a ton of workarounds we might need to do.

clabby:
@mattsse nw. We'll add a manual check in the proof that disallows use of the 7702 tx. gl w/ alphanet launch 🫡

PR #168, opened 2024-10-14
feat(envelope): add missing `From<Signed<TxEip7702>>`

hai-rise:
## Motivation

We use this pattern downstream (cc @kien-rise):  
https://github.com/risechain/pevm/blob/bed6f220c343d03e6bef0dc7540c531444450d57/src/chain/optimism.rs#L85-L88 

Which doesn't work for `TxEip7702`.

## Solution

Add the missing `From<Signed<TxEip7702>>` for completeness.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -125,6 +125,12 @@ impl From<Signed<TxEip1559>> for OpTxEnvelope {
     }
 }
 
+impl From<Signed<TxEip7702>> for OpTxEnvelope {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
ty,

heads up I had to add 7702 support here to, but I think we want to transition to a standalone type for odyssey-alloy instead and likely remove this again shortly

hai-rise:
@mattsse No worries. We live on the (cutting) edge together 🫡!

PR #169, opened 2024-10-15
eip1559 execution payload

cody-wang-cb:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (commented):
I believe this is no longer relevant now that eip1559 data is encoded in the extra data

PR #171, opened 2024-10-15
feat: add depositTransaction trait

hoank101:
This PR will help this issue https://github.com/paradigmxyz/reth/issues/11736
 converted into a trait from reth to alloy-rs

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/consensus/src/transaction/deposit.rs
```diff
@@ -241,6 +241,31 @@ impl Decodable for TxDeposit {
     }
 }
 
+pub trait DepositTransaction {
+    fn source_hash(&self) -> Option<B256>;
+    fn mint(&self) -> Option<u128>;
+    fn is_system_transaction(&self) -> bool;
+    fn is_deposit(&self) -> bool;
+}
```
mattsse, line 249: this needs some docs

mattsse, line 249: I'd also like to move this to a new file `src/traits.rs`

mattsse (changes requested):

crates/consensus/src/transaction/deposit.rs
hoank101: thank @mattsse, j just update, please help me review it!

emhane (approved):

refcell (approved):

mattsse (approved):
@refcell context is here that we need this as prep for moving reth off of optimism features

PR #172, opened 2024-10-15
feat(protocol): Batch

refcell:
### Description

`kona-derive` is nearly decoupled from protocol-level types. The final remaining types are the batch types: single and span batches.

This PR ports over all batch types from `kona-derive`'s batch module, introducing a new minimal trait abstraction for batch validity checks called the `BatchValidator`. Effectively, it is a subset of the `L2ChainProvider` trait in `kona-derive`.

### Additional Context

Works on https://github.com/anton-rs/kona/issues/695

### DIFF
--- a/crates/protocol/src/traits.rs
+++ b/crates/protocol/src/traits.rs
@@ -1,0 +1,25 @@
+//! Traits for working with protocol types.
+
+use alloc::{boxed::Box, string::ToString};
+use async_trait::async_trait;
+use core::fmt::Display;
+use op_alloy_consensus::OpBlock;
+
+use crate::L2BlockInfo;
+
+/// Describes the functionality of a data source that fetches safe blocks.
+#[async_trait]
+pub trait BatchValidator {
+    /// The error type for the [BatchValidator].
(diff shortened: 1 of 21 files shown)


crates/protocol/src/traits.rs
```diff
@@ -0,0 +1,25 @@
...
+use alloc::{boxed::Box, string::ToString};
+use async_trait::async_trait;
+use core::fmt::Display;
+use op_alloy_consensus::OpBlock;
+
+use crate::L2BlockInfo;
+
+/// Describes the functionality of a data source that fetches safe blocks.
+#[async_trait]
+pub trait BatchValidator {
```
clabby, line 12: Slight preference for renaming this as `BatchValidator` is a stage in the pipeline now. Maybe `BatchValidationProvider`?

refcell: I like this, will open a fresh stack... this pr is gonna be too heinous to review.

refcell:
This PR is replaced by the stack started in #186

PR #175, opened 2024-10-17
chore: bump alloy

yash-atreya:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #176, opened 2024-10-18
chore: bump alloy and remove `OpExecutionPayloadV4`

onbjerg:
## Motivation

There is no longer a V4 execution payload in Pectra, so it does not make sense that there would be one for Optimism. The v4 payload envelope still exists, but it now wraps a v3 payload.

See also https://github.com/alloy-rs/alloy/pull/1383

Based on https://github.com/alloy-rs/op-alloy/pull/175

## Solution

Remove the V4 payload.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,18 +47,18 @@ alloy-rlp = { version = "0.3", default-features = false }
 alloy-primitives = { version = "0.8", default-features = false }
 alloy-sol-types = { version = "0.8", default-features = false }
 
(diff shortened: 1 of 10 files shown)


mattsse (approved):
we assume for now that this will also be the case for op

mattsse:
pending alloy release

mattsse (approved):

PR #177, opened 2024-10-19
chore: Dependency Updates

refcell:
### Description

Dependency updates


crates/rpc-types/src/lib.rs
```diff
@@ -5,7 +5,7 @@
 )]
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
-#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(not(any(test, feature = "std")), no_std)]
```
mattsse, line 8: we can actually remove this entirely, let me do that real quick

mattsse (approved):

PR #186, opened 2024-10-25
feat(protocol): Batch Type

refcell:
### Description


crates/protocol/src/batch/type.rs
```diff
@@ -0,0 +1,77 @@
...
+            _ => panic!("Invalid batch type: {val}"),
+        }
+    }
+}
+
+impl From<&[u8]> for BatchType {
+    fn from(buf: &[u8]) -> Self {
+        Self::from(buf[0])
+    }
+}
```
onbjerg, line 45: is this used anywhere? imo no point given there's `From<u8>`, so this should be handled where its used instead

onbjerg (approved):
smol comment otherwise lgtm

PR #187, opened 2024-10-25
feat(protocol): Batch Validity

refcell:
### Description

Part of a port migrating the batch types from `kona-derive` to `op-alloy`. See [`kona#695`](https://github.com/anton-rs/kona/issues/695).

This PR introduces the `BatchValidity` type.

PR #188, opened 2024-10-25
feat(protocol): SingleBatch Type

refcell:
### Description

This PR introduces the `SingleBatch` type.


refcell:
## 📚 $\text{Stack Overview}$

* #188 👈
* #187
* #186
* `main`

crates/protocol/src/batch/single.rs
```diff
@@ -0,0 +1,246 @@
...
+            return if cfg.is_holocene_active(inclusion_block.timestamp) {
+                BatchValidity::Past
+            } else {
+                BatchValidity::Drop
+            };
+        }
+        BatchValidity::Accept
+    }
+
+    /// Checks if the batch is valid.
```
onbjerg, line 61: nit: would be great to link to the specs if there are up to date ones

crates/protocol/src/batch/single.rs
```diff
@@ -0,0 +1,246 @@
...
+                return BatchValidity::Drop;
+            }
+            return BatchValidity::Future;
+        }
+        if self.timestamp < next_timestamp {
+            return if cfg.is_holocene_active(inclusion_block.timestamp) {
+                BatchValidity::Past
+            } else {
+                BatchValidity::Drop
+            };
```
onbjerg, line 56: nit: can we pick either `return if` or `if .. { return }`

onbjerg (approved):
otherwise lgtm

crates/protocol/src/batch/single.rs
refcell: Great suggestion.

PR #190, opened 2024-10-27
feat(protocol): Batch Error Types

refcell:
### Description

Ports batch error types to `op-alloy-protocol`.

### Provenance

Part of a port migrating the batch types from `kona-derive` to `op-alloy`. See [`kona#695`](https://github.com/anton-rs/kona/issues/695).

PR #191, opened 2024-10-27
feat(protocol): Batch Utilities

refcell:


refcell:
## 📚 $\text{Stack Overview}$

* #191 👈
* #190
* `main`

crates/protocol/src/utils.rs
```diff
@@ -224,9 +231,85 @@ fn u24(input: &[u8], idx: u32) -> u32 {
         + (u32::from(input[(idx + 2) as usize]) << 16)
 }
 
+/// Reads transaction data from a reader.
+#[allow(unused)]
```
onbjerg, line 235: I don't think we need these `#[allow]`s since they are marked as `pub`?

crates/protocol/src/utils.rs
```diff
@@ -224,9 +231,85 @@ fn u24(input: &[u8], idx: u32) -> u32 {
...
+pub fn read_tx_data(r: &mut &[u8]) -> Result<(Vec<u8>, TxType), SpanBatchError> {
+    let mut tx_data = Vec::new();
+    let first_byte =
+        *r.first().ok_or(SpanBatchError::Decoding(SpanDecodingError::InvalidTransactionData))?;
```
onbjerg, line 239: we can just use `Buf::get_u8()` and similar here instead of manually doing `.first()` and `advance`.

crates/protocol/src/utils.rs
```diff
@@ -224,9 +231,85 @@ fn u24(input: &[u8], idx: u32) -> u32 {
...
+                Ok((v - 35) & 1 == 1)
+            } else {
+                // Unprotected legacy txs must have v = 27 or 28
+                Ok(v - 27 == 1)
+            }
+        }
+        TxType::Eip2930 | TxType::Eip1559 => Ok(v == 1),
+        _ => Err(SpanBatchError::Decoding(SpanDecodingError::InvalidTransactionType)),
+    }
+}
```
onbjerg, line 288: this entire fn can be replaced with [`Parity::y_parity`](https://docs.rs/alloy-primitives/latest/alloy_primitives/enum.Parity.html#method.y_parity) at call sites:

crates/protocol/src/utils.rs
refcell: I don't think so since we only advance if the tx type is legacy, otherwise we need to leave the first byte to decode the header. Maybe I can put up a test to demonstrate this

crates/protocol/src/utils.rs
```diff
@@ -224,18 +231,112 @@ fn u24(input: &[u8], idx: u32) -> u32 {
...
+            let v = tx.signature().v().to_u64();
+            if 64 - v.leading_zeros() <= 8 {
+                return v != 27 && v != 28 && v != 1 && v != 0;
+            }
+            // anything not 27 or 28 is considered protected
+            true
+        }
+        _ => true,
+    }
+}
```
refcell, line 284: FYI @onbjerg, I've upstreamed this to `TxEnvelope` in alloy: https://github.com/alloy-rs/alloy/pull/1578

refcell: We can refactor this out later once it lands in a release

clabby (approved):

PR #192, opened 2024-10-28
feat(protocol): Span Batch Signature Type

refcell:


refcell:
## 📚 $\text{Stack Overview}$

* #192 👈
* #191
* #190
* `main`

onbjerg (commented):
what is the difference between this and [`Signature`](https://docs.rs/alloy-primitives/latest/alloy_primitives/struct.Signature.html)?

refcell:
> what is the difference between this and [`Signature`](https://docs.rs/alloy-primitives/latest/alloy_primitives/struct.Signature.html)?

I think we can remove this and just use [`Signature`](https://docs.rs/alloy-primitives/latest/alloy_primitives/struct.Signature.htm).

PR #194, opened 2024-10-28
feat(protocol): Span Batch Bits

refcell:
### Description

Ports the span batch bits to `op-alloy-protocol`.

### Provenance

### DIFF
--- a/crates/protocol/src/batch/bits.rs
+++ b/crates/protocol/src/batch/bits.rs
@@ -1,0 +1,224 @@
+//! Module for working with span batch bits.
+
+use crate::SpanBatchError;
+use alloc::{vec, vec::Vec};
+use alloy_rlp::Buf;
+use core::cmp::Ordering;
+
+/// Type for span batch bits.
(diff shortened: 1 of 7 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #194 👈
* #193
* #192
* #191
* #190
* `main`

crates/protocol/src/batch/bits.rs
```diff
@@ -0,0 +1,224 @@
+//! Module for working with span batch bits.
+
+use crate::SpanBatchError;
+use alloc::{vec, vec::Vec};
+use alloy_rlp::Buf;
+use core::cmp::Ordering;
+
+/// Type for span batch bits.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct SpanBatchBits(pub Vec<u8>);
```
onbjerg, line 10: Since we have `get_bit` and `set_bit`, I don't think having the inner `Vec<u8>` marked `pub` makes sense

onbjerg (approved):
smol nit otherwise lgtm

PR #198, opened 2024-10-28
feat(protocol): Span Batch Validity Checks

refcell:
### Description

Ports span batch validity checks to `op-alloy-protocol`.

### Provenance


refcell:
> [!NOTE]
> Stacked ontop of #197

PR #200, opened 2024-10-28
feat(protocol): Batch

refcell:
### Description

Ports the remaining batch types to `op-alloy-protocol`.

### Provenance

Last PR in the stack to migrate batch types from `kona-derive` to `op-alloy`. See [`kona#695`](https://github.com/anton-rs/kona/issues/695).


crates/protocol/src/batch/errors.rs
```diff
@@ -35,6 +35,41 @@ impl core::error::Error for SpanBatchError {
...
+impl From<SpanBatchError> for BatchDecodingError {
+    fn from(err: SpanBatchError) -> Self {
+        Self::SpanBatchError(err)
+    }
+}
+
+impl core::error::Error for BatchDecodingError {
+    fn source(&self) -> Option<&(dyn core::error::Error + 'static)> {
+        match self {
+            Self::SpanBatchError(err) => Some(err),
```
clabby, line 67: Can't use `alloy_rlp::Error` here? Does it not impl `core::error::Error`?

clabby (approved):

crates/protocol/src/batch/errors.rs
refcell: Yep, does not impl core error

PR #203, opened 2024-10-30
Rename `TxDeposit::signature` to `pre_bedrock_signature`

emhane:
## Motivation

Fixes semantics of `TxDeposit` to differentiate between ovm and evm chain

## Solution

Renames `TxDeposit::signature` to `pre_bedrock_signature`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/deposit.rs
+++ b/crates/consensus/src/transaction/deposit.rs
@@ -174,9 +174,9 @@ impl TxDeposit {
         }
     }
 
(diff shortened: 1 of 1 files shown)


mattsse (commented):
hmm we also use this signature post bedrock for things like rpc I believe?

mattsse:
yeah this is actually included by all deposit transactions after bedrock, so we can keep it as is

PR #205, opened 2024-10-30
[wip] feat: bump alloy

klkvr:
## Motivation

Bumps alloy, alloy-core and alloy-eip7702.

Code changes: 
- https://github.com/alloy-rs/alloy/pull/1496 changed transaction encoding API, encoding functions are now prefixed with `rlp`,`eip2718` or `network`. I've updated `TxDeposit` to have methods following this pattern
- https://github.com/alloy-rs/core/pull/776 updated signature type to avoid using `Parity` generic and instead just keep a boolean for parity byte. The motivation is to have more correctness in encoding and data structure. Before that it was possible to construct `Signed<TxLegacy>` where transaction chain_id is `Some`, but the parity would not follow EIP-155.
    
   This required a few changes to `SpanBatchTransactions` type:
     - `encode_y_parity_bits` and `encode_tx_sigs_rs` are merged into single `encode_tx_sigs` which firstly encodes parity bits and then `rs` values. Same was done for decoding methods
     - `recover_v` method is removed. instead, we now pass the `is_protected` flag down to `SpanBatchLegacyTransactionData` and only set `chain_id` to `Some` when `is_protected` is true

Blocked by minor releases for alloy, core and revm

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,8 +61,8 @@ alloy-network-primitives = { version = "0.5.4", default-features = false }
 alloy-rlp = { version = "0.3", default-features = false }
 
 # Alloy Core
-alloy-sol-types = { version = "0.8", default-features = false }
-alloy-primitives = { version = "0.8", default-features = false }
+alloy-sol-types = { version = "0.8.10", default-features = false }
+alloy-primitives = { version = "0.8.10", default-features = false }
 
 # Revm
 revm = "17.1.0"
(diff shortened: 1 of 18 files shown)


klkvr:
I've removed `test_span_batch_transactions_add_invalid_legacy_parity_decoding` test cc @refcell. Unsure what exactly it was asserting but with this PR legacy tx parity is always correctly encoded

Cargo.toml
```diff
@@ -116,15 +116,19 @@ rstest = "0.23.0"
...
+alloy-consensus = { git = "https://github.com/alloy-rs/alloy", rev = "d99f09c" }
+alloy-network = { git = "https://github.com/alloy-rs/alloy", rev = "d99f09c" }
+alloy-network-primitives = { git = "https://github.com/alloy-rs/alloy", rev = "d99f09c" }
+alloy-rpc-types-eth = { git = "https://github.com/alloy-rs/alloy", rev = "d99f09c" }
+alloy-rpc-types-engine = { git = "https://github.com/alloy-rs/alloy", rev = "d99f09c" }
+alloy-eips = { git = "https://github.com/alloy-rs/alloy", rev = "d99f09c" }
+alloy-serde = { git = "https://github.com/alloy-rs/alloy", rev = "d99f09c" }
+alloy-signer = { git = "https://github.com/alloy-rs/alloy", rev = "d99f09c" }
+alloy-provider = { git = "https://github.com/alloy-rs/alloy", rev = "d99f09c" }
+alloy-transport = { git = "https://github.com/alloy-rs/alloy", rev = "d99f09c" }
```
refcell, line 134: Calling this out in a comment so we don't forget to remove these patches.

refcell (approved):
This is very nice work @klkvr

mattsse (approved):

PR #215, opened 2024-11-01
feat: `OpTransactionRequest`

klkvr:
## Motivation

Closes https://github.com/alloy-rs/op-alloy/issues/27

To be able to replace `UnsingedTx` and `TxEnvelope` on Optimism network with op-consensus types, we need a op-specific tx builder.

## Solution

Add `OpTransactionBuilder` which is a wrapper around `TransactionBuilder`. It intentionally does not have any direct methods allowing to set EIP-4844 fields, however they can still be set through `AsMut` or during deserialization.

When building typed transaction, eip-4844 transactions are getting converted to eip-1559

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/Cargo.toml
+++ b/crates/network/Cargo.toml
@@ -24,3 +24,4 @@ alloy-consensus.workspace = true
 alloy-network.workspace = true
 alloy-primitives.workspace = true
 alloy-rpc-types-eth.workspace = true
+alloy-signer.workspace = true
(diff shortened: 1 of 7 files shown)


mattsse (approved):

klkvr:
Removing op-alloy-rpc-types crate from no-std checks, will re-add in #205 which will have https://github.com/alloy-rs/alloy/pull/1602

PR #234, opened 2024-11-06
feat(protocol): Batch Reader

refcell:
### Description

Introduces the batch reader type.

Fixes broken batch decoding.


refcell:
## 📚 $\text{Stack Overview}$

* #234 👈
* `main`

crates/protocol/src/batch/transactions.rs
```diff
@@ -260,11 +260,8 @@ impl SpanBatchTransactions {
                 .tx_sigs
                 .get(idx as usize)
                 .ok_or(SpanBatchError::Decoding(SpanDecodingError::InvalidTransactionData))?;
-            let is_protected = self
-                .protected_bits
-                .get_bit(idx as usize)
-                .ok_or(SpanBatchError::Decoding(SpanDecodingError::InvalidTransactionData))?
-                == 1;
+            let is_protected =
+                self.protected_bits.get_bit(idx as usize).unwrap_or(sig.v() as u8) == 1;
```
refcell, line 264: @klkvr in your alloy update pr, span batch txs decoding broke when a tx didn't have protected bits - I've updated the `SpanBatchTransactions` handling here to use the signature v value if the protected bits are missing. This fixes batch reading. I've upstreamed the `BatchReader` from kona as well in this PR which validates batch reading - prior to this pr, decoding a valid batch with the `BatchReader` failed.

refcell: For reference: https://github.com/anton-rs/kona/pull/786

klkvr: looks like before `None` was treated as false already

https://github.com/alloy-rs/op-alloy/blob/edc82479bc6b0d2f48176f278638eaedf892b5f9/crates/protocol/src/batch/transactions.rs#L247-L250

crates/protocol/src/batch/reader.rs
```diff
@@ -0,0 +1,128 @@
+//! Contains the [BatchReader] which is used to iteratively consume batches from raw data.
+
+use alloc::vec::Vec;
+use alloy_primitives::Bytes;
+use alloy_rlp::Decodable;
+use miniz_oxide::inflate::decompress_to_vec_zlib;
+use op_alloy_genesis::RollupConfig;
+
```
clabby, line 8:
```suggestion
```

crates/protocol/src/batch/transactions.rs
clabby: Yeah if the protected bit isn't found, we should be treating the signature as unprotected.

clabby (approved):
LGTM, pending protected bit fix.

crates/protocol/src/brotli.rs
```diff
@@ -0,0 +1,105 @@
+//! Stage Utilities
+
+use alloc::{vec, vec::Vec};
+use alloc_no_stdlib::*;
```
mattsse, line 4: this seems weird, why do we need this?

mattsse (approved):
lgtm, but as discussed lets get rid of this dep

https://github.com/alloy-rs/op-alloy/blob/aac1566ed65015fbabafe5ba0bb49def6af3012d/crates/rpc-types-engine/Cargo.toml#L19-L19

refcell:
Will come back to this once more batch type cleaning is done

refcell:
Closing this for now - batch reading will strictly be used in the derivation pipeline so it shouldn't need to be used in other op stack components.

PR #259, opened 2024-11-15
feat(protocol): Batch Encoding

refcell:
### Description

Implements encoding for batch types.

Steps in the direction to batch submission.

### DIFF
--- a/crates/protocol/src/batch/core.rs
+++ b/crates/protocol/src/batch/core.rs
@@ -1,7 +1,9 @@
 //! Module containing the core [Batch] enum.
 
-use crate::{BatchDecodingError, BatchType, RawSpanBatch, SingleBatch, SpanBatch};
-use alloy_rlp::{Buf, Decodable};
+use crate::{
+    BatchDecodingError, BatchEncodingError, BatchType, RawSpanBatch, SingleBatch, SpanBatch,
+};
+use alloy_rlp::{Buf, Decodable, Encodable};
 use op_alloy_genesis::RollupConfig;
 
 /// A Batch.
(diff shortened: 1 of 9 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #259 👈
* `main`

crates/protocol/src/batch/core.rs
```diff
@@ -48,4 +51,82 @@ impl Batch {
             }
         }
     }
+
+    /// Attempts to encode the batch to a writer.
+    pub fn encode(&self, out: &mut Vec<u8>) -> Result<(), BatchEncodingError> {
```
mattsse, line 56: you could also use `out: &mut dyn bytes::BufMut` here then this accepts vec and bytesmut
but vec is fine

crates/protocol/src/batch/raw.rs
```diff
@@ -16,10 +16,10 @@ pub struct RawSpanBatch {
     pub payload: SpanBatchPayload,
 }
 
-impl TryFrom<SpanBatch> for RawSpanBatch {
+impl TryFrom<&SpanBatch> for RawSpanBatch {
```
mattsse, line 19: TryFrom references are usually horrbile to work with, recommend adding function to`SpanBatch` as well `try_to_raw(&self` or smth

mattsse (approved):

crates/protocol/src/batch/raw.rs
refcell: Yea good callout. Will follow up with this!

PR #260, opened 2024-11-15
fix(book): Batch over SingleBatch

refcell:
### Description

Since #259 adds `Batch` encoding, book examples can decode and encode directly to and from the `Batch` types instead of the `SingleBatch` type.


refcell:
## 📚 $\text{Stack Overview}$

* #260 👈
* #259
* `main`

PR #274, opened 2024-11-16
chore(consensus): Re-export and Hardfork Cleanup

refcell:
### Description

Cleans up re-exports in `op-alloy-consensus`.

Moves towards a world where addresses can be customizable for `Hardforks`.
Would like to see the `AddressList` come into play here in there future for extensibility.
Should also update `Hardforks` to provide a more extensible pattern to add future hardforks.

Might be worth renaming `Hardforks` to "Protocol Upgrades" to be in-line with the OP Stack [specs](https://specs.optimism.io).


refcell:
## 📚 $\text{Stack Overview}$

* #274 👈
* `main`

mattsse (approved):

PR #282, opened 2024-11-17
chore(consensus): Move OpTxType and add tests

refcell:
### Description

Moves `OpTxType` from the `envelope` mod into it's own `tx_type` mod to make it more digestible.

Also adds tests, including an `OpTxType::ALL` test that requires a newly-added variant is added to the list.


refcell:
## 📚 $\text{Stack Overview}$

* #282 👈
* `main`

Issue #85, opened 2024-09-09
[Feature] Use `U256` instead of `u128` for `mint`


yash-atreya:
### Component

rpc-types

### Describe the feature you would like

              wondering if `mint` should be `U256` instead in op-alloy as this is basically same as `value`

_Originally posted by @klkvr in https://github.com/foundry-rs/foundry/pull/8595#discussion_r1703587404_

Make `mint` field `U256` in the following places

1. https://github.com/alloy-rs/op-alloy/blob/1c0c8bae27e49d688e4caa3e4ab08bc1973e1a5a/crates/consensus/src/transaction/deposit.rs#L26

2. https://github.com/alloy-rs/op-alloy/blob/1c0c8bae27e49d688e4caa3e4ab08bc1973e1a5a/crates/rpc-types/src/transaction.rs#L18

programskillforverification:
It should be `U256` in op-alloy

reymom:
Is this still open?

mahmudsudo:
Hi , can i take on this issue ?

klkvr:
@emhane I think this is still useful given that `value` is U256 everywhere?

PR #286, opened 2024-11-18
feat: Use U256 instead fo u128 for mint

DougieHauser:
Contribution proposal for [issue 85](https://github.com/alloy-rs/op-alloy/issues/85)

Using U256 for mint instead of u128


mattsse:
@refcell I assume u128 is sufficient here?

I believe this change could be problematic for reth

tynes:
> @refcell I assume u128 is sufficient here?
> 
> I believe this change could be problematic for reth

If somebody somehow is able to actually deposit that much ether in a single transaction it would cause a chain split without this change. Unlikely that somebody would be able to do that tho

PR #289, opened 2024-11-19
chore(consensus): Trait Abstracted Hardforks

refcell:
### Description

Trait abstracts hardforks to make them more ergonomic to extend, modify, or introduce a new hardfork.


crates/consensus/src/hardforks/traits.rs
```diff
@@ -0,0 +1,9 @@
+//! The trait abstraction for a Hardfork.
+
+use alloy_primitives::Bytes;
+
+/// The trait abstraction for a Hardfork.
+pub trait Hardfork {
+    /// Returns the hardfork upgrade transactions as [Bytes].
+    fn txs(&self) -> impl Iterator<Item = Bytes> + '_;
+}
```
emhane, line 9: how about returning the `TxDeposit` instead and letting the caller encode it

mattsse (changes requested):
for debugging purposes, could be useful to have an iterator over txenvelope and the trait function does a map (encode2718)

crates/consensus/src/hardforks/traits.rs
mattsse: the caller is likely only interested in the raw txs, but it could be useful to have an additional fn that returns tx envelopes

emhane: would be nice with an alloy repo for rollup abstractions. this trait is likely relevant for other evm rollups, but not for l1 alloy.

refcell: Pushed the changes - let me know if this is what you're looking for :)

PR #295, opened 2024-11-20
chore: add default for txtype

mattsse:
doesn't hurt, need this because default enforced in reth for testing


refcell (approved):

refcell:
Was actually thinking about adding this a bit ago but wasn't sure if it'd be worthwhile or useful for reth

PR #296, opened 2024-11-21
fix(protocol): Remove panic in brotli compress method

refcell:
### Description

Removes an unintended panic in the brotli compression method.


refcell:
## 📚 $\text{Stack Overview}$

* #296 👈
* `main`

crates/consensus/src/hardforks/ecotone.rs
```diff
@@ -79,118 +89,86 @@ impl super::Hardforks {
...
-        .encode_2718(&mut buffer);
-        txs.push(Bytes::from(buffer));
+            },
+        ])
+        .into_iter()
+    }
+
+    /// Returns the list of [OpTxEnvelope]s for the Ecotone network upgrade.
+    pub fn enveloped_txs() -> impl Iterator<Item = OpTxEnvelope> {
+        Self::deposits().map(|deposit| OpTxEnvelope::from(deposit.seal_slow()))
```
mattsse, line 161: seems a bit redundant to hash these when we only need them encoded?

crates/consensus/src/hardforks/forks.rs
```diff
@@ -0,0 +1,62 @@
...
+/// ```
+///
+/// Build fjord hardfork upgrade transactions:
+/// ```rust
+/// use op_alloy_consensus::{Hardfork, Hardforks};
+/// let fjord_upgrade_txs = Hardforks::FJORD.txs();
+/// assert_eq!(fjord_upgrade_txs.collect::<Vec<_>>().len(), 3);
+/// ```
+#[derive(Debug, Default, Clone, Copy)]
+pub struct Hardforks;
```
mattsse, line 28:
```suggestion
#[non_exhaustive]
pub struct Hardforks;
```

mattsse (approved):
lgtm, one suggestion

crates/consensus/src/hardforks/ecotone.rs
```diff
@@ -79,118 +89,86 @@ impl super::Hardforks {
...
+    }
+}
 
-        txs
+impl super::Hardfork for Ecotone {
+    /// Constructs the Ecotone network upgrade transactions.
+    fn txs(&self) -> impl Iterator<Item = Bytes> + '_ {
+        Self::enveloped_txs().map(|tx| {
```
mattsse, line 168: this should probably use deposits direclty?

refcell: We need to encode the `OpTxEnvelope` wrapper type which internally hashes `TxDeposit` in it's [`From<TxDeposit> for OpTxEnvelope`](https://github.com/alloy-rs/op-alloy/blob/main/crates/consensus/src/transaction/envelope.rs#L71-L75) impl.

refcell: I've added tests to assert the bytes. If you try to use `TxDeposit` itself, with rlp encoding, it doesn't match the desired bytes.

refcell: Let me know if I'm misunderstanding but I think you meant using the `deposits` method directly e.g.

```rust
https://github.com/alloy-rs/op-alloy/blob/main/crates/consensus/src/transaction/envelope.rs#L71-L75
```

mattsse: this is the wrong encode fn, you need 2718?

see how encodable2718 is impl for optxenvelope

mattsse:
@refcell for txdeposit we can actually move all of those into the trait impl Encode/Decode2718

https://github.com/alloy-rs/op-alloy/blob/0959cbfb727b1fe05d8ff52b85faab11fe8ef160/crates/consensus/src/transaction/deposit.rs#L189-L193

mattsse (approved):

crates/consensus/src/hardforks/ecotone.rs
refcell: Ah I missed the 2718 methods. Great! I removed and we can follow up with another pr to move 2718 methods into the `Encodable2718/Decodable2718` trait impls.

PR #298, opened 2024-11-21
chore(protocol): Move and Extend Brotli Compression

refcell:
### Description

Moves brotli compression to the correct `brotli` module.

Extends the brotli compression method to accept the brotli compression level.

See: https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/types.go#L50

PR #299, opened 2024-11-21
feat(protocol): Compressors

refcell:
### Description

Introduces compressors into `op-alloy-protocol`, cleaning up the brotli and zlib utility methods as well as the `ChannelOut` type.


refcell:
## 📚 $\text{Stack Overview}$

* #299 👈
* #298
* #296
* `main`

crates/protocol/src/channel_out.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// The encoded batch exceeds the max RLP bytes per channel.
+    #[error("The encoded batch exceeds the max RLP bytes per channel")]
+    ExceedsMaxRlpBytesPerChannel,
+}
+
+/// [ChannelOut] constructs a channel from compressed, encoded batch data.
+#[derive(Debug, Clone)]
+pub struct ChannelOut<'a, C>
+where
+    C: Compressor + Clone + core::fmt::Debug,
```
mattsse, line 38: I think you can relax this here entirely

crates/protocol/src/channel_out.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub frame_number: u16,
+    /// Compressed batch data.
+    pub compressed: Option<Bytes>,
+    /// The compressor.
+    pub compressor: C,
+}
+
+impl<'a, C> ChannelOut<'a, C>
+where
+    C: Compressor + Clone + core::fmt::Debug,
```
mattsse, line 60: is clone + debug required here?

crates/protocol/src/compression/zlib.rs
```diff
@@ -0,0 +1,28 @@
...
+use crate::Compressor;
+use alloc::vec::Vec;
+use miniz_oxide::inflate::DecompressError;
+
+/// The best compression.
+const BEST_ZLIB_COMPRESSION: u8 = 9;
+
+/// The ZLIB compressor.
+#[derive(Debug, Clone)]
+pub struct ZlibCompressor;
```
mattsse, line 12:
```suggestion
#[derive(Debug, Clone, Default)]
#[non_exhaustive]
pub struct ZlibCompressor;
```

mattsse (approved):
lgtm

crates/protocol/src/channel_out.rs
refcell: I think so in order for the struct to be `Clone + Debug` since it's composition

PR #308, opened 2024-11-23
chore(registry): Dogfood Test Rollup Config

refcell:
### Description

Dogfoods the hardcoded test rollup config and uses the new `From<&RollupConfig> for HardForkConfiguration`.

### DIFF
--- a/crates/genesis/src/chain.rs
+++ b/crates/genesis/src/chain.rs
@@ -55,6 +55,19 @@ pub struct HardForkConfiguration {
     pub holocene_time: Option<u64>,
 }
 
+impl From<&RollupConfig> for HardForkConfiguration {
+    fn from(config: &RollupConfig) -> Self {
+        Self {
+            canyon_time: config.canyon_time,
+            delta_time: config.delta_time,
+            ecotone_time: config.ecotone_time,
+            fjord_time: config.fjord_time,
+            granite_time: config.granite_time,
+            holocene_time: config.holocene_time,
+        }
+    }
+}
+
 /// Defines core blockchain settings per block.
 ///
 /// Tailors unique settings for each network based on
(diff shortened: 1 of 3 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #308 👈
* `main`

crates/genesis/src/chain.rs
```diff
@@ -55,6 +55,19 @@ pub struct HardForkConfiguration {
     pub holocene_time: Option<u64>,
 }
 
+impl From<&RollupConfig> for HardForkConfiguration {
```
mattsse, line 58: these types are in the same crate.

this impl would be less horrible if we also add this natively to RollupConfig.

I guess `from<&>` wouldn't be an issue for `RollupConfig` because we assume this will likely be Arc'ed.
but still, very much dislike standalone from ref impls

crates/registry/src/superchain.rs
```diff
@@ -138,38 +134,14 @@ mod tests {
...
             chain: String::new(),
-            hardfork_configuration: HardForkConfiguration {
-                canyon_time: Some(1704992401),
-                delta_time: Some(1708560000),
-                ecotone_time: Some(1710374401),
-                fjord_time: Some(1720627201),
-                granite_time: Some(1726070401),
-                holocene_time: None,
-            },
+            hardfork_configuration: (&crate::configs::BASE_MAINNET_CONFIG).into(),
```
mattsse, line 137: see, this is horrible,

mattsse, line 137:
```suggestion
            hardfork_configuration: crate::configs::BASE_MAINNET_CONFIG.hardfork_config(),
```

is much nicer

mattsse (changes requested):

mattsse (approved):

PR #314, opened 2024-11-26
feat: bump alloy

royvardhan:
Ref: https://github.com/alloy-rs/alloy/pull/1672#pullrequestreview-2457745263


royvardhan:
@mattsse What do you think?

crates/consensus/src/receipt/receipts.rs
```diff
@@ -105,15 +121,20 @@ impl OpTxReceipt for OpDepositReceipt {
 #[derive(Clone, Debug, PartialEq, Eq, Default)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
-pub struct OpDepositReceiptWithBloom<T = Log> {
+pub struct OpDepositReceiptWithBloom<T = OpDepositReceipt<Log>> {
```
mattsse, line 124: @klkvr do we even need this now, because this should be identical with the alloy receiptWithBloom type @klkvr ?

and we could add an alias instead?

klkvr: yeah, this could be `type OpDepositReceiptWithBloom<T = Log> = ReceiptWithBloom<OpDepositReceipt<T>>` now

mattsse (approved):
lgtm, pending alloy release

PR #319, opened 2024-11-30
chore: use non_exhaustive attribute

mattsse:


mattsse:
actually non_exhaustive still allows instantiations

PR #325, opened 2024-12-03
feat: add miner extension trait

mattsse:
ref 

https://github.com/paradigmxyz/reth/issues/13092

and

https://github.com/ethereum-optimism/op-geth/blob/0a46245ccc5c801e7b18c258aceb5327d8ad69ad/eth/api_miner.go#L60-L62

### DIFF
--- a/crates/rpc-jsonrpsee/src/traits.rs
+++ b/crates/rpc-jsonrpsee/src/traits.rs
@@ -136,14 +136,12 @@ pub trait EngineApiExt {
     async fn signal_superchain_v1(&self, signal: SuperchainSignal) -> RpcResult<ProtocolVersion>;
 }
 
-
 /// Op API extension for controlling the miner.
 #[cfg_attr(not(feature = "client"), rpc(server, namespace = "miner"))]
 #[cfg_attr(feature = "client", rpc(server, client, namespace = "miner"))]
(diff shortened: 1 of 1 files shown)


crates/rpc-jsonrpsee/src/traits.rs
```diff
@@ -135,3 +135,13 @@ pub trait EngineApiExt {
     #[method(name = "signalSuperchainV1")]
     async fn signal_superchain_v1(&self, signal: SuperchainSignal) -> RpcResult<ProtocolVersion>;
 }
+
+/// Op API extension for controlling the miner.
+#[cfg_attr(not(feature = "client"), rpc(server, namespace = "miner"))]
+#[cfg_attr(feature = "client", rpc(server, client, namespace = "miner"))]
```
refcell, line 141: Curious, is it destructive to do the following instead?
```rust
#[rpc(server, namespace = "miner")]
#[cfg_attr(feature = "client", rpc(client))]
```

refcell (approved):

crates/rpc-jsonrpsee/src/traits.rs
mattsse: iirc this didn't quite work, but can't remember specifics

PR #330, opened 2024-12-03
Propagate arbitrary

emhane:
Based on https://github.com/alloy-rs/op-alloy/pull/329

## Motivation

Incomplete arbitrary feature propagation

## Solution

Enables `alloy-primitives/arbitrary` in local `arbtirary` feature

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
we could also install zepter which is pretty convenient for feature mgmt

emhane:
> we could also install zepter which is pretty convenient for feature mgmt

100%, was thinking to open an issue for precisely this

PR #305, opened 2024-11-22
chore(protocol): Compressor Implementations

refcell:
### Description

Re-works compression to use compression writers that allows for streaming compression.

PR #335, opened 2024-12-05
feat(protocol): Compressors with Mocked Brotli Streaming

refcell:
### Description

Introduces various compressors for batch submission.

This replaces #305 for now since the brotli compressor writer isn't playing nice.

Streaming compression is instead mocked by using the earlier `compress_brotli` method with a wrapper that implements the `CompressorWriter` trait.


crates/protocol/src/channel_out.rs
```diff
@@ -32,10 +31,10 @@ pub enum ChannelOutError {
 }
 
 /// [ChannelOut] constructs a channel from compressed, encoded batch data.
-#[derive(Debug, Clone)]
+#[allow(missing_debug_implementations)]
 pub struct ChannelOut<'a, C>
 where
-    C: Compressor + Clone + core::fmt::Debug,
+    C: ChannelCompressor,
```
mattsse, line 37: looks like this can be relaxed

crates/protocol/src/compression/variant.rs
```diff
@@ -1,15 +1,17 @@
...
-/// implementing the [Compressor] trait itself.
+/// implementing the [ChannelCompressor] trait itself.
 #[derive(Debug, Clone)]
 pub enum VariantCompressor {
     /// The brotli compressor.
-    Brotli(BrotliCompressor),
+    Brotli(Box<BrotliCompressor>),
     /// The zlib compressor.
     Zlib(ZlibCompressor),
 }
```
mattsse, line 17: compressor trait looks object safe, so I assume `Box<dyn Compressor>` should also work

crates/protocol/src/channel_out.rs
refcell: wdym, should I add back the `Clone + core::fmt::Debug` restriction?

crates/protocol/src/compression/variant.rs
refcell: The `Box` wrapper was legacy due to the streaming compression writer impl. That will be a follow-up pr

crates/protocol/src/channel_out.rs
mattsse: you can get rid of this where clause here entirely because it's not necessary for the struct def itself

mattsse (changes requested):

crates/protocol/src/compression/ratio.rs
```diff
@@ -4,6 +4,77 @@
 //!
 //! [rc]: https://github.com/ethereum-optimism/optimism/blob/develop/op-batcher/compressor/ratio_compressor.go#L7
```
tynes, line 5: you should use a permalink here to a particular commit, otherwise it will go stale

mattsse (approved):

PR #341, opened 2024-12-14
feat: add OpPooledTransaction

Rjected:
## Motivation

We need a type that represents a pooled transaction, this cannot include deposit transactions

## Solution

Implemented `OpPooledTransaction`, which is like `OpTxEnvelope` but without the deposit variant.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


gakonst:
isn't this the same as the normal pooled tx?

mattsse (approved):
lgtm, added a conversion for the OpTxenvelope variant

mattsse:
> isn't this the same as the normal pooled tx?

no 4844, it's also not the same as the consensus format, because we'd like to exclude deposit variants

PR #352, opened 2025-01-02
feat(protocol): Interop Types

refcell:
### Description

Adds interop message types to `op-alloy-protocol` following the [ExecutingMessage spec](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/predeploys.md#executingmessage-event)
and the associated [op-geth code](https://github.com/ethereum-optimism/op-geth/blob/optimism/core/types/interoptypes/interop.go).

### DIFF
--- a/crates/protocol/src/messages/events.rs
+++ b/crates/protocol/src/messages/events.rs
@@ -35,7 +35,7 @@ sol! {
     ) external;
 }
 
-/// A [MessagePayload] is the raw payload of an initiating message.
+/// A [`MessagePayload`] is the raw payload of an initiating message.
 #[derive(Debug, Clone, From, AsRef, PartialEq, Eq)]
 pub struct MessagePayload(Bytes);
 
@@ -50,7 +50,7 @@ impl From<Log> for MessagePayload {
     }
 }
 
-/// A [MessageIdentifier] uniquely represents a log that is emitted from a chain within
+/// A [`MessageIdentifier`] uniquely represents a log that is emitted from a chain within
 /// the broader dependency set. It is included in the calldata of a transaction sent to the
 /// CrossL2Inbox contract.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
@@ -69,7 +69,7 @@ pub struct MessageIdentifier {
 }
 
 impl MessageIdentifier {
-    /// Decode a [MessageIdentifier] from ABI-encoded data.
+    /// Decode a [`MessageIdentifier`] from ABI-encoded data.
     pub fn abi_decode(data: &[u8], validate: bool) -> Result<Self, alloy_sol_types::Error> {
         MessageIdentifierAbi::abi_decode(data, validate).map(|abi| abi.into())
     }
--- a/crates/protocol/src/messages/events.rs
+++ b/crates/protocol/src/messages/events.rs
@@ -1,0 +1,129 @@
+//! Message event primitives for OP stack interoperability.
+//!
+//! <https://specs.optimism.io/interop/messaging.html#messaging>
+//! <https://github.com/ethereum-optimism/optimism/blob/34d5f66ade24bd1f3ce4ce7c0a6cfc1a6540eca1/packages/contracts-bedrock/src/L2/CrossL2Inbox.sol>
+use alloc::{vec, vec::Vec};
+use alloy_primitives::{keccak256, Address, Bytes, Log, U256};
+use alloy_sol_types::{sol, SolType};
+
+sol! {
+    /// @notice The struct for a pointer to a message payload in a remote (or local) chain.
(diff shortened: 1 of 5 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #352 👈
* `main`

crates/protocol/src/messages/events.rs
```diff
@@ -0,0 +1,129 @@
...
+    fn from(payload: MessagePayload) -> Self {
+        payload.0
+    }
+}
+
+impl AsRef<[u8]> for MessagePayload {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
```
emhane, line 74: prefer if you can remove as much of this boiler plate code as possible deriving it instead
https://docs.rs/derive_more/latest/derive_more/all.html

crates/protocol/src/messages/events.rs
```diff
@@ -0,0 +1,129 @@
...
+
+/// A [MessageIdentifier] uniquely represents a log that is emitted from a chain within
```
emhane, line 76:
```suggestion
/// A [`MessageIdentifier`] uniquely represents a log that is emitted from a chain within
```
let's adhere to standard of code formatting types/traits/modules, see for example https://doc.rust-lang.org/std/alloc/index.html

emhane (changes requested):
first pass

emhane (commented):
makes sense to include this type in scope
https://github.com/ethereum-optimism/op-geth/blob/58aed221675c46e09e004ebae24a8bcb95d4cbfe/core/types/interoptypes/interop.go#L21-L24

crates/protocol/src/messages/events.rs
```diff
@@ -0,0 +1,129 @@
...
+/// the broader dependency set. It is included in the calldata of a transaction sent to the
+/// CrossL2Inbox contract.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct MessageIdentifier {
```
mattsse, line 81: should these use camelcase?

refcell:
> makes sense to include this type in scope https://github.com/ethereum-optimism/op-geth/blob/58aed221675c46e09e004ebae24a8bcb95d4cbfe/core/types/interoptypes/interop.go#L21-L24

I believe the sol macro generates this for the `ExecutingMessage` which we define [here](https://github.com/alloy-rs/op-alloy/pull/352/files#diff-bdea6c7bac984616b2562e8a8e4f239c000f11b02f71f2138ce63c607b10dae2R21-R25).

Think we could use the generated `ExecutingMessage` but not opposed to clearly defining a custom struct.

crates/protocol/src/messages/events.rs
refcell: Yes good catch

crates/protocol/src/messages/events.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// The account that sent the message.
+    pub origin: Address,
+    /// The block number that the message was sent in.
+    pub block_number: u64,
+    /// The log index of the message in the block (global).
+    pub log_index: u64,
+    /// The timestamp of the message.
+    pub timestamp: u64,
+    /// The chain ID of the chain that the message was sent on.
+    #[cfg_attr(feature = "serde", serde(rename = "chainID"))]
```
tynes, line 69: not sure if this is `chainId` or `chainID`

emhane (approved):
ok, I follow with the `ExecutingMessage`, thanks! that's what I need to use then ion my pr stacked onto this one :)

crates/protocol/src/messages/events.rs
refcell: `chainID`, see [op-geth](https://github.com/ethereum-optimism/op-geth/blob/optimism/core/types/interoptypes/interop.go#L101)

crates/protocol/src/messages/events.rs
```diff
@@ -0,0 +1,135 @@
...
+            chainId: U256::from(id.chain_id),
+        }
+    }
+}
+
+impl From<executeMessageCall> for ExecutingMessage {
+    fn from(call: executeMessageCall) -> Self {
+        Self { id: call._id, msgHash: keccak256(call._message.as_ref()) }
+    }
+}
```
emhane, line 108: very nice! so this is equivalent to https://github.com/ethereum-optimism/op-geth/blob/58aed221675c46e09e004ebae24a8bcb95d4cbfe/core/types/interoptypes/interop.go#L26-L68, capisci

clabby (approved):

PR #364, opened 2025-01-08
chore(protocol): migrate op-alloy-protocol->op-rs/maili-protocol

emhane:
Migrates `op-alloy-protocol`->`op-rs/maili-protocol`. Re-exports `maili-protocol::*`.


refcell (approved):
LGTM

PR #366, opened 2025-01-08
chore(registry): Migrate `op-alloy-registry`->`op-rs/maili-registry`

emhane:
Migrates `op-alloy-registry` to `op-rs/maili-registry`. Re-exports `maili_registry::*`.


crates/registry/etc/bind.sh
emhane, line 1: have a hunch that I need to keep this folder in order to not break libraries using `op-alloy` as dep..
same goes for the empty superchain-registry folder and the Justfile. is that right @refcell ?

emhane: if you're unsure I'll leave it for now and check with in the op-reth development lab fork

refcell: Oh interesting... I don't think so because the json files are read as bytes by the registry. The bind script is only for updating those json files - technically it's not even used at build time

crates/protocol/src/test_utils.rs
refcell, line 1: I think a rebase might remove these?

crates/protocol/src/utils.rs
refcell, line 1: ibid

crates/registry/etc/bind.sh
emhane: right and the chain list is used in constructing `maili_registry::Registry`

crates/protocol/src/utils.rs
emhane: ibid 😂 imma start using this in github too

crates/registry/etc/bind.sh
emhane: and how is the empty superchain-registry folder used?

refcell (approved):

crates/registry/etc/bind.sh
emhane: has been resolved
context: git reads superchain-registry, fetched online, into folder

PR #367, opened 2025-01-08
chore(rpc-types-engine): Migrate `op_alloy_rpc_types_engine::superchain`->`op-rs/maili-rpc-types-engine`

emhane:
Migrates `op_alloy_rpc_types_engine::superchain` to `op-rs/maili-rpc-types-engine`. Re-exports `maili::rpc_types_engine::*`.


Cargo.toml
```diff
@@ -68,6 +68,7 @@ alloy-primitives = { version = "0.8.12", default-features = false }
 # Maili
 maili-provider = { version = "0.1.0", default-features = false }
 maili-protocol = { version = "0.1.0", default-features = false }
+maili = { version = "0.1.0", default-features = false, features = ["rpc-types-engine"] }
```
refcell, line 71: I think we can use [`maili-rpc-types-engine`](https://crates.io/crates/maili-rpc-types-engine)

refcell (approved):

PR #373, opened 2025-01-09
chore(provider): Remove Provider Crate

refcell:
### Description

The provider crate only contains op specific types that don't extend or relate to ethereum types.

This PR removes `op-alloy-provider` since `maili-provider` can be used in place.


refcell:
## 📚 $\text{Stack Overview}$

* #373 👈
* `main`

mattsse (approved):
makes sense

PR #377, opened 2025-01-10
chore(consensus): Migrate deposit source to `maili-common`

emhane:
Migrates deposit source types to `maili-common`


crates/rpc-jsonrpsee/src/traits.rs
```diff
@@ -0,0 +1,162 @@
...
+//! Rollup Node
+
+use alloc::{boxed::Box, string::String, vec::Vec};
+use core::net::IpAddr;
+
+use alloy_eips::BlockNumberOrTag;
+use alloy_primitives::{B256, U64};
+use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+use maili_common::{ProtocolVersion, SuperchainSignal};
+use maili_protocol::{ExecutingMessage, SafetyLevel};
```
mattsse, line 12: I think we should consider moving those traits to maili as well, because I'd prefer if the repo dependency is one directional

emhane: yes we are working on removing the dep on op-alloy in maili

refcell (approved):

PR #386, opened 2025-01-14
fix(consensus): Remove re-export of `maili` traits

emhane:
Removes re-export of `maili` deposit traits, to avoid error 
`note: there are multiple different versions of crate `maili_common` in the dependency graph`


refcell (approved):

emhane:
> > note: there are multiple different versions of crate maili_common in the dependency graph
> 
> this is because we still have a circular dep graph across the repo I'd really prefer if we could get rid of this

yepp, aware of this. we are almost done getting rid of last dep `op-alloy-consensus`. apologies for the construction site.

Issue #398, opened 2025-01-22
[Feature] Revert maili migration


emhane:
### Component

other

### Describe the feature you would like

Revert maili migration to the extent that op-alloy doesn't depend on maili

PR #399, opened 2025-01-22
chore: revert maili deps

mattsse:
Closes https://github.com/alloy-rs/op-alloy/issues/398

this reverts a bunch of stuff and removes all maili dependencies.

this is definitely incomplete but is the first step towards restoring op-alloy as a maili independent repo.

the dep graph looks now as follows:
alloy -> op-alloy -> maili

maili can have all types limited to protocol impls (although debatable if this is actually beneficial, imo this just slows down development atm).

next steps are invert deps on maili side and clean things up there.

this repo would benefit from the flz stuff, so I suggest we move this crate in here and only keep the brotli stuff in maili

### DIFF
--- a/crates/rpc-jsonrpsee/src/traits.rs
+++ b/crates/rpc-jsonrpsee/src/traits.rs
@@ -1,107 +1,9 @@
 #![allow(missing_docs)]
 
-//! Rollup Node
+//! Various `jsonrpsee` docs
 
-use alloc::{boxed::Box, string::String, vec::Vec};
-use core::net::IpAddr;
-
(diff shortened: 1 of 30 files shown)


refcell:
Happy to deprecate maili-flz crate once we move it in here.

crates/rpc-jsonrpsee/src/traits.rs
```diff
@@ -0,0 +1,34 @@
+#![allow(missing_docs)]
```
refcell, line 1: I think we can remove this - most items look like they have docs

refcell (approved):

PR #400, opened 2025-01-22
feat: add flz

mattsse:
adds standalone flz repo

this could eventually be converted into a standalone crate, if revm is willing to import that and remove code dup


clabby:
LGTM.

We'd also like to have `revm-optimism` (which is a standalone crate, post-refactor on revm's main branch) consume maili for the L1Block type. What do you think about that?

We discussed with Dragan last week, and he said it was fine, so long as we moved op-revm outside of the revm repository.

clabby (approved):

mattsse:
you we faced some opposition getting standalone alloy deps included in revm, if this would be a standalone thing then this would be easier

PR #401, opened 2025-01-24
chore: rm execution requests from v4 payload fn

### LINKED ISSUE #395: [Feature] Check empty EL reqs `newPayloadV4`
### Component

provider

Check that `execution_requests` are always empty
https://github.com/alloy-rs/op-alloy/blob/b506e16d6e6bb0f3cf416c965092b693fea4cb71/crates/provider/src/ext/engine.rs#L210-L225

Update docs accordingly
https://github.com/alloy-rs/op-alloy/blob/b506e16d6e6bb0f3cf416c965092b693fea4cb71/crates/provider/src/ext/engine.rs#L52

### Additional context

https://github.com/ethereum-optimism/specs/issues/506

mattsse:
closes #395

### DIFF
--- a/crates/provider/src/ext/engine.rs
+++ b/crates/provider/src/ext/engine.rs
@@ -54,7 +54,6 @@ pub trait OpEngineApi<N, T> {
         &self,
         payload: ExecutionPayloadV3,
         parent_beacon_block_root: B256,
-        execution_requests: Vec<Bytes>,
     ) -> TransportResult<PayloadStatus>;
 
     /// Updates the execution layer client with the given fork choice, as specified for the Shanghai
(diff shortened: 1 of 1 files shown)


emhane:
conflicts with spec https://github.com/ethereum-optimism/specs/pull/525

emhane:
hm, though this isn't an rpc api...so it could be ok to remove the arg...in rpc it breaks spec though

mattsse:
> - `executionRequests` MUST be an empty array.

what am I missing here?

meyer9:
Since this is op-specific, I think it makes sense to remove the param from the public interface (but obviously still send it over RPC). This LGTM!

Issue #407, opened 2025-02-03
[Feature] Define `OpExecutionPayloadSidecar` and isthmus payload fields


emhane:
### Component

consensus

### Describe the feature you would like

Define `OpExecutionPayloadSidecar` which should contain cancun fields and isthmus fields, where isthmus fields consists of storage root of `L2ToL1MessagePasser.sol`, aka l2 withdrawals root. Look at how it's done for alloy equivalent `ExecutionPayloadSidecar`.

PR #410, opened 2025-02-03
feat(isthmus): define `IsthmusPayloadFields`

emhane:
Closes https://github.com/alloy-rs/op-alloy/issues/407


crates/rpc-types-engine/src/sidecar.rs
```diff
@@ -0,0 +1,102 @@
...
+/// Container type for all available additional `newPayload` request parameters that are not present
+/// in the [`ExecutionPayload`](alloy_rpc_types_engine::ExecutionPayload) object itself.
+///
+/// Default is equivalent to pre-cancun, payloads v1 and v2.
+#[derive(Debug, Clone, Default)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct OpExecutionPayloadSidecar {
+    /// Cancun request params introduced in `engine_newPayloadV3` that are not present in the
+    /// [`ExecutionPayload`](alloy_rpc_types_engine::ExecutionPayload).
+    cancun: MaybeCancunPayloadFields,
```
mattsse, line 15: should we add a note here that this is actually Canyon ? but equivalent to L1 cancun?

crates/rpc-types-engine/src/sidecar.rs
```diff
@@ -0,0 +1,102 @@
...
+
+/// A container type for [`IsthmusPayloadFields`] that may or may not be present.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, From, Into)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[from(IsthmusPayloadFields)]
+pub struct MaybeIsthmusPayloadFields {
+    fields: Option<IsthmusPayloadFields>,
+}
+
+impl OpExecutionPayloadSidecar {
```
mattsse, line 40: can we move thos below the struct def

crates/rpc-types-engine/src/sidecar.rs
```diff
@@ -0,0 +1,102 @@
...
+        self.isthmus.fields.as_ref()
+    }
+
+    /// Consumes the type and returns the [`IsthmusPayloadFields`].
+    pub fn into_isthmus(self) -> Option<IsthmusPayloadFields> {
+        self.isthmus.into()
+    }
+
+    /// Returns the parent beacon block root, if any.
+    pub fn parent_beacon_block_root(&self) -> Option<B256> {
```
mattsse, line 99: probably want a fn for withdrawals root as well?

crates/rpc-types-engine/src/sidecar.rs
```diff
@@ -0,0 +1,102 @@
...
+    isthmus: MaybeIsthmusPayloadFields,
+}
+
+/// Fields introduced in `engine_newPayloadV4` that are not present in the
+/// [`ExecutionPayload`](alloy_rpc_types_engine::ExecutionPayload) RPC object.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, Constructor)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct IsthmusPayloadFields {
+    /// EIP-7685 requests.
+    pub withdrawals_root: B256,
```
mattsse, line 29: since this is just one field, that can be mocked, we don't necessarily need this type, and could use option directly, but imo having this makes it consistent and easier to document

mattsse (approved):
lgtm,

smol nits

crates/rpc-types-engine/src/sidecar.rs
emhane: was wondering if it more had smthg to do with serde serialisation, hence copied `MaybePayloadFields` pattern from alloy

mattsse (approved):

Issue #415, opened 2025-02-04
[Feature] Add enum OpExecutionPayload


mattsse:
### Component

rpc-types-engine

### Describe the feature you would like

same as L1 `pub enum ExecutionPayload{v1,2,3}` but with op's variants, which now has a v4 variant:

https://github.com/alloy-rs/op-alloy/pull/414

emhane:
except v1

PR #416, opened 2025-02-04
Define `OpExecutionPayload`

emhane:
Closes https://github.com/alloy-rs/op-alloy/issues/415


crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+#[cfg_attr(feature = "serde", serde(untagged))]
+pub enum OpExecutionPayload {
+    /// V2 payload
+    V2(ExecutionPayloadV2),
+    /// V3 payload
+    V3(ExecutionPayloadV3),
+    /// V4 payload
+    V4(OpExecutionPayloadV4),
```
mattsse, line 20: lgtm,

one thing that could be problematic, perhaps even for L1 types is that untagged here doesn't quite work for V4 because this could already succeed on v3, 
for the engineapi this isn't problematic, but we should track manual deserde for this

mattsse (approved):
lgtm,

next we need https://github.com/alloy-rs/op-alloy/issues/417

Issue #417, opened 2025-02-04
[Feature] Add OpExecutionData


mattsse:
### Component

rpc-types-engine

### Describe the feature you would like

same as https://github.com/alloy-rs/alloy/pull/2003 but with Op types

PR #418, opened 2025-02-04
feat(isthmus): define `OpExecutionData`

emhane:
Closes https://github.com/alloy-rs/op-alloy/issues/417

- Defines `OpExecutionData` which wraps `OpExecutionPayload` and `OpExecutionPayloadSidecar`
- Debugs `OpExecutionPayload`
- Removes `OpExecutionPayloadV4`, since the l2 withdrawals root goes in sidecar, it doesn't add an extra field to the header

### DIFF
--- a/crates/rpc-types-engine/src/envelope.rs
+++ b/crates/rpc-types-engine/src/envelope.rs
@@ -6,6 +6,19 @@
 use alloy_primitives::{keccak256, PrimitiveSignature as Signature, B256};
 use alloy_rpc_types_engine::ExecutionPayload;
 
+use crate::{OpExecutionPayload, OpExecutionPayloadSidecar};
+
+/// Struct aggregating [`OpExecutionPayload`] and [`OpExecutionPayloadSidecar`] and encapsulating
+/// complete payload supplied for execution.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct OpExecutionData {
+    /// Execution payload.
+    pub payload: OpExecutionPayload,
+    /// Additional fork-specific fields.
+    pub sidecar: OpExecutionPayloadSidecar,
+}
+
 /// Optimism execution payload envelope in network format.
 ///
 /// This struct is used to represent payloads that are sent over the Optimism
--- a/crates/rpc-types-engine/src/payload/mod.rs
+++ b/crates/rpc-types-engine/src/payload/mod.rs
@@ -9,7 +9,7 @@ use alloy_rpc_types_engine::{ExecutionPayloadV2, ExecutionPayloadV3};
 /// An execution payload, which can be either [`ExecutionPayloadV2`], [`ExecutionPayloadV3`], or
 /// [`OpExecutionPayloadV4`].
 #[derive(Clone, Debug, PartialEq, Eq)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(untagged))]
 pub enum OpExecutionPayload {
     /// V2 payload
--- a/crates/rpc-types-engine/src/lib.rs
+++ b/crates/rpc-types-engine/src/lib.rs
@@ -15,7 +15,7 @@ mod attributes;
 pub use attributes::OpPayloadAttributes;
 
 mod envelope;
-pub use envelope::{OpNetworkPayloadEnvelope, PayloadEnvelopeError, PayloadHash};
+pub use envelope::{OpExecutionData, OpNetworkPayloadEnvelope, PayloadEnvelopeError, PayloadHash};
 
 mod sidecar;
 pub use sidecar::{IsthmusPayloadFields, MaybeIsthmusPayloadFields, OpExecutionPayloadSidecar};


crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -3,19 +3,201 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, From)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(untagged))]
 pub enum OpExecutionPayload {
     /// V2 payload
     V2(ExecutionPayloadV2),
     /// V3 payload
     V3(ExecutionPayloadV3),
-    /// V4 payload
-    V4(OpExecutionPayloadV4),
```
mattsse, line 20: why is this removed?

mattsse (commented):
> Removes OpExecutionPayloadV4, since the l2 withdrawals root goes in sidecar, it doesn't add an extra field to the header

I'm very confused by this now, 
so the newPayloadV4 endpoint gets an additional argument and the payload object remains unchanged?

crates/rpc-types-engine/src/payload/mod.rs
emhane: ye

emhane: see https://specs.optimism.io/protocol/isthmus/exec-engine.html#update-to-executabledata

mattsse: still unclear what the parameters are, from reading the spec I'd assume that the withdrawals root is an additional attribute of the payload param and not a standalone parameter?

emhane: yes, we know, we are working on it

crates/rpc-types-engine/src/payload/v4.rs
```diff
@@ -4,21 +4,6 @@ use alloc::vec::Vec;
...
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
-pub struct OpExecutionPayloadV4 {
-    /// L1 execution payload
-    #[cfg_attr(feature = "serde", serde(flatten))]
-    pub payload_inner: ExecutionPayloadV3,
-    /// OP-Stack Isthmus specific field:
-    /// instead of computing the root from a withdrawals list, set it directly.
-    /// The "withdrawals" list attribute must be non-nil but empty.
-    pub withdrawals_root: B256,
```
mattsse, line 19: how are we getting this field over rpc now?

mattsse (commented):
this is a bit confusing, how does the intended flow look like from rpc pov?

the isthmuspayloadfields(withdrawalsroot) actually doesn't make much sense because we need to capture this as part of the payload over rpc.

so are we doing
parse OpPayloadV4 on newPayload, split then split that into V3 and Isthmus payloadfields?

crates/rpc-types-engine/src/envelope.rs
```diff
@@ -24,32 +74,7 @@ pub struct OpNetworkPayloadEnvelope {
 
 impl OpNetworkPayloadEnvelope {
     /// Decode a payload envelope from a snappy-compressed byte array.
-    /// The payload version decoded is `ExecutionPayloadV1` from SSZ bytes.
-    #[cfg(feature = "std")]
-    pub fn decode_v1(data: &[u8]) -> Result<Self, PayloadEnvelopeError> {
```
refcell, line 29: Dumb Q - but why was this removed?

refcell (commented):
Would it make sense to split the changes requested in #417 into it's own PR as [the upstream alloy pr did](https://github.com/alloy-rs/alloy/pull/2003) for ethereum types?

This way we can split out the endpoint changes which is causing confusion.

mattsse:
> type into it's own PR without modifying other code

makes sense,

I'll summarize the core issue with this PR:

* with isthmus opstack breaks L1 compatibility with the payload parameter of the `newPayloadV4([payload, (sidecar)])` endpoint.
* because we need to parse `payload` as an object, we need a dedicated opstack compatible ExecutionPayload type (this type is currently named `OpExecutionPayloadV4`)
* We currently have `IsthumusPayloadFields` for the withdrawals_root but this is already part of the payload object, so this type seems redundant

The problem from reth POV
* because newPayload does not accepts multiple param objects, not a single object we need to add the helper type `ExecutionData` so that we can easily handle arbitrary execution requests, this is effectively extra work necessary due to engine API design

crates/rpc-types-engine/src/payload/v4.rs
emhane: always as envelope afaik

crates/rpc-types-engine/src/envelope.rs
emhane: because op doesn't support v1 https://specs.optimism.io/protocol/exec-engine.html can be out of scope to fix this semantic in this pr

crates/rpc-types-engine/src/payload/v4.rs
mattsse: the envelope is what `getPayload` returns, this type is the raw payload

I know this is confusing

emhane:
> Would it make sense to split the changes requested in #417 into it's own PR as [the upstream alloy pr did](https://github.com/alloy-rs/alloy/pull/2003) for ethereum types? That is, specifically split the `OpExecutionData` type into it's own PR without modifying other code.
> 
> This way we can split out the endpoint changes which is causing confusion.

done, removed the impl body of the new type that i had implemented to match the alloy type `ExecutionData` impl, since this was what required me to change parts of existing code

emhane:
> > type into it's own PR without modifying other code
> 
> makes sense,
> 
> I'll summarize the core issue with this PR:
> 
> * with isthmus opstack breaks L1 compatibility with the payload parameter of the `newPayloadV4([payload, (sidecar)])` endpoint.
> * because we need to parse `payload` as an object, we need a dedicated opstack compatible ExecutionPayload type (this type is currently named `OpExecutionPayloadV4`)
> * We currently have `IsthumusPayloadFields` for the withdrawals_root but this is already part of the payload object, so this type seems redundant
> 
> The problem from reth POV
> 
> * because newPayload does not accepts multiple param objects, not a single object we need to add the helper type `ExecutionData` so that we can easily handle arbitrary execution requests, this is effectively extra work necessary due to engine API design

all great points that I'm glad surface from this pr. makes sense to open a separate issues.

mattsse (approved):

PR #427, opened 2025-02-11
feat: add additional compat impls

mattsse:
blocked by https://github.com/alloy-rs/alloy/pull/2042

adds some conversion helpers for txdepoist

### DIFF
--- a/crates/consensus/Cargo.toml
+++ b/crates/consensus/Cargo.toml
@@ -21,7 +21,9 @@ alloy-eips.workspace = true
 alloy-consensus.workspace = true
 alloy-primitives = { workspace = true, features = ["rlp"] }
 
-# Maili
+# compat
+alloy-network = { workspace = true, optional = true }
+alloy-rpc-types-eth = { workspace = true, optional = true }
 
 # misc
 thiserror.workspace = true
@@ -51,6 +53,7 @@ std = [
     "alloy-consensus/std",
     "derive_more/std",
 ]
+alloy-compat = ["dep:alloy-network", "dep:alloy-rpc-types-eth"]
 k256 = ["alloy-primitives/k256", "alloy-consensus/k256"]
 kzg = ["alloy-eips/kzg", "alloy-consensus/kzg", "std"]
 arbitrary = [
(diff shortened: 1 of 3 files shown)


cakevm:
I am really looking forward to this change. I wanted to let you know that I have created a minimal example for testing here: https://github.com/cakevm/alloy-to-reth-primitives/blob/main/src/main.rs The idea here is that it makes testing ExEx easier by fetching blocks from RPC and ingesting them into Reth. Like the benchmark in reth is doing.

PS: Maybe it makes sense to get rid of some feature flags from `reth-optimism-primitives` to be on par with the `reth-ethereum-primitives`crate.

mattsse:
> Maybe it makes sense to get rid of some feature flags from reth-optimism-primitives to be on par with the reth-ethereum-primitivescrate.

such as?
could you perhaps summarize this in a reth issue?

mattsse:
@cakevm 
this https://github.com/cakevm/alloy-to-reth-primitives/blob/main/src/main.rs is helpful

this should become easier: https://github.com/alloy-rs/alloy/issues/2033

cakevm:
> > Maybe it makes sense to get rid of some feature flags from reth-optimism-primitives to be on par with the reth-ethereum-primitivescrate.
> 
> such as? could you perhaps summarize this in a reth issue?

It is really a minor thing. Maybe I am also missing something, but it feels odd that I require different features for the same functionality. But probably I miss some other dependencies why this is required. When I have I bigger picture I can open a issue of course.

As you can see in the example:
```toml
# eth-reth
reth-ethereum-primitives = { git = "https://github.com/paradigmxyz/reth", rev = "1e0b0d8", features = [
  "reth-codec"
] }

# op-reth
reth-optimism-primitives = { git = "https://github.com/paradigmxyz/reth", rev = "1e0b0d8", features = [
  "alloy-compat",
  "optimism",
  "reth-codec",
  "serde",
] }
```

mattsse:
there's currently a limitation where some reth crates are incompatible dues to revm optimism features, this will go away within the next few weeks and then we can combine all of this

Issue #421, opened 2025-02-07
[Feature] Impl `OpExecutionData` behaviour


emhane:
### Component

rpc-types-engine

### Describe the feature you would like

Impl `OpExecutionData` behaviour. Refer to behaviour of alloy type `ExecutionData`.

### Additional context

blocked by https://github.com/alloy-rs/op-alloy/issues/417

varun-doshi:
@emhane we would also need `impl OpExecutionPayload` for this

varun-doshi:
I can take this one

emhane:
closed in https://github.com/alloy-rs/op-alloy/pull/429

PR #429, opened 2025-02-11
feat: impl OpExecutionData

varun-doshi:
## Motivation

Ref #421 

## Solution

Implemented some functions for `OpExecutionData` similar to `ExecutionData` from alloy

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-engine/Cargo.toml
+++ b/crates/rpc-types-engine/Cargo.toml
@@ -41,6 +41,7 @@ derive_more = { workspace = true, default-features = false, features = [
   "into",
 ]}
 
+
 [dev-dependencies]
 arbtest.workspace = true
 serde_json.workspace = true
--- a/crates/rpc-types-engine/src/envelope.rs
+++ b/crates/rpc-types-engine/src/envelope.rs
@@ -19,6 +19,28 @@ pub struct OpExecutionData {
     pub sidecar: OpExecutionPayloadSidecar,
 }
 
+impl OpExecutionData {
+    /// Creates new instance of [`ExecutionData`].
+    pub const fn new(payload: OpExecutionPayload, sidecar: OpExecutionPayloadSidecar) -> Self {
+        Self { payload, sidecar }
+    }
+
+    /// Returns the parent hash of the block.
+    pub fn parent_hash(&self) -> B256 {
+        self.payload.parent_hash()
+    }
+
+    /// Returns the hash of the block.
+    pub fn block_hash(&self) -> B256 {
+        self.payload.block_hash()
+    }
+
+    /// Returns the number of the block.
+    pub fn block_number(&self) -> u64 {
+        self.payload.block_number()
+    }
+}
+
 /// Optimism execution payload envelope in network format.
 ///
 /// This struct is used to represent payloads that are sent over the Optimism
--- a/crates/rpc-types-engine/src/envelope.rs
+++ b/crates/rpc-types-engine/src/envelope.rs
@@ -20,7 +20,7 @@ pub struct OpExecutionData {
 }
 
 impl OpExecutionData {
-    /// Creates new instance of [`ExecutionData`].
+    /// Creates new instance of [`OpExecutionData`].
     pub const fn new(payload: OpExecutionPayload, sidecar: OpExecutionPayloadSidecar) -> Self {
         Self { payload, sidecar }
     }
--- a/crates/rpc-types-engine/src/envelope.rs
+++ b/crates/rpc-types-engine/src/envelope.rs
@@ -26,17 +26,17 @@ impl OpExecutionData {
     }
 
     /// Returns the parent hash of the block.
-    pub fn parent_hash(&self) -> B256 {
+    pub const fn parent_hash(&self) -> B256 {
         self.payload.parent_hash()
     }
 
     /// Returns the hash of the block.
-    pub fn block_hash(&self) -> B256 {
+    pub const fn block_hash(&self) -> B256 {
         self.payload.block_hash()
     }
 
     /// Returns the number of the block.
-    pub fn block_number(&self) -> u64 {
+    pub const fn block_number(&self) -> u64 {
         self.payload.block_number()
     }
 }
--- a/crates/rpc-types-engine/src/payload/mod.rs
+++ b/crates/rpc-types-engine/src/payload/mod.rs
@@ -84,7 +84,7 @@ impl OpExecutionPayload {
         match self {
             Self::V2(payload) => payload.try_into_block(),
             Self::V3(payload) => payload.try_into_block(),
-            Self::V4(payload) => payload.payload_inner.try_into_block(),
+            Self::V4(payload) => payload.try_into_block(),
         }
     }
 }
--- a/crates/rpc-types-engine/src/payload/mod.rs
+++ b/crates/rpc-types-engine/src/payload/mod.rs
@@ -73,10 +73,11 @@ impl OpExecutionPayload {
         self.as_v2().payload_inner.block_number
     }
 
+    #[allow(rustdoc::broken_intra_doc_links)]
     /// Converts [`OpExecutionPayload`] to [`Block`].
     ///
     /// Caution: This does not set fields that are not part of the payload and only part of the
-    /// [`OpExecutionPayloadSidecar`]:
+    /// [`OpExecutionPayloadSidecar`](crate::OpExecutionPayloadSidecar):
     /// - parent_beacon_block_root
     ///
     /// See also: [`OpExecutionPayload::try_into_block_with_sidecar`]
--- a/crates/rpc-types-engine/src/payload/mod.rs
+++ b/crates/rpc-types-engine/src/payload/mod.rs
@@ -44,8 +44,8 @@ impl OpExecutionPayload {
     /// Returns a reference to the V3 payload, if any.
     pub const fn as_v3(&self) -> Option<&ExecutionPayloadV3> {
         match self {
-            Self::V2(payload) => None,
-            Self::V3(payload) => Some(&payload),
+            Self::V2(_) => None,
+            Self::V3(payload) => Some(payload),
             Self::V4(payload) => Some(&payload.payload_inner),
         }
     }
@@ -53,7 +53,7 @@ impl OpExecutionPayload {
     /// Returns a mutable reference to the V3 payload, if any.
     pub fn as_v3_mut(&mut self) -> Option<&ExecutionPayloadV3> {
         match self {
-            Self::V2(payload) => None,
+            Self::V2(_) => None,
             Self::V3(payload) => Some(payload),
             Self::V4(payload) => Some(&payload.payload_inner),
         }
--- a/crates/rpc-types-engine/src/payload/mod.rs
+++ b/crates/rpc-types-engine/src/payload/mod.rs
@@ -3,8 +3,8 @@
 pub mod v3;
 pub mod v4;
 
-use crate::{OpExecutionPayloadSidecar, OpExecutionPayloadV4};
-use alloy_consensus::{Block, EMPTY_ROOT_HASH};
+use crate::OpExecutionPayloadV4;
+use alloy_consensus::Block;
 use alloy_eips::Decodable2718;
 use alloy_primitives::B256;
 use alloy_rpc_types_engine::{ExecutionPayloadV2, ExecutionPayloadV3, PayloadError};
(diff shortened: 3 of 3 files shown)


varun-doshi:
This is a starting point...can work up from here

mattsse (commented):
oh @emhane 
don't we have that already?

https://github.com/alloy-rs/op-alloy/blob/0061ee7b593066d52c0229247f3e55ed3a033924/crates/rpc-types-engine/src/envelope.rs#L15-L15

my bad for not closing the issue?

emhane:
yes but I removed the behaviour I impl to shrink scope @mattsse

crates/rpc-types-engine/src/envelope.rs
```diff
@@ -19,6 +19,28 @@ pub struct OpExecutionData {
     pub sidecar: OpExecutionPayloadSidecar,
 }
 
+impl OpExecutionData {
+    /// Creates new instance of [`ExecutionData`].
```
emhane, line 23:
```suggestion
    /// Creates new instance of [`OpExecutionData`].
```

mattsse:
ah I see

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -19,3 +22,53 @@ pub enum OpExecutionPayload {
...
+
+impl OpExecutionPayload {
+    /// Returns a reference to the V2 payload, if any.
+    pub const fn as_v2(&self) -> Option<&ExecutionPayloadV2> {
+        match self {
+            Self::V2(payload) => Some(payload),
+            Self::V3(payload) => Some(&payload.payload_inner),
+            Self::V4(payload) => todo!(),
+        }
+    }
```
emhane, line 34:
```suggestion
    /// Returns a reference to the V2 payload, if any.
    pub const fn as_v2(&self) -> &ExecutionPayloadV2 {
        match self {
            Self::V2(payload) => payload,
            Self::V3(payload) => &payload.payload_inner,
            Self::V4(payload) => todo!(),
        }
    }
```
we always have v2

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -19,3 +22,53 @@ pub enum OpExecutionPayload {
...
+
+    /// Returns a mutable reference to the V2 payload, if any.
+    pub fn as_v2_mut(&mut self) -> Option<&mut ExecutionPayloadV2> {
```
emhane, line 37:
```suggestion
    pub fn as_v2_mut(&mut self) -> &mut ExecutionPayloadV2 {
```
same here, we always have v2

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -19,3 +22,53 @@ pub enum OpExecutionPayload {
...
+    pub const fn block_hash(&self) -> B256 {
+        self.as_v2().unwrap().payload_inner.block_hash
+    }
+
+    /// Returns the block number for this payload.
+    pub const fn block_number(&self) -> u64 {
+        self.as_v2().unwrap().payload_inner.block_number
+    }
+
+    /// Converts [`ExecutionPayloadV1`] to [`Block`].
```
emhane, line 59:
```suggestion
    /// Converts [`OpExecutionPayload`] to [`Block`].
```

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -19,3 +22,53 @@ pub enum OpExecutionPayload {
...
+    ///
+    /// Caution: This does not set fields that are not part of the payload and only part of the
+    /// [`ExecutionPayloadSidecar`]:
+    /// - parent_beacon_block_root
+    /// - requests_hash
```
emhane, line 64:
```suggestion
```
always empty for OP, so we aren't concerned with it here

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -19,3 +22,53 @@ pub enum OpExecutionPayload {
...
+    ///
+    /// See also: [`ExecutionPayload::try_into_block_with_sidecar`]
```
emhane, line 66:
```suggestion
    /// See also: [`OpExecutionPayload::try_into_block_with_sidecar`]
```
we need this method to be impl as well as update docs here

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -19,3 +22,53 @@ pub enum OpExecutionPayload {
...
+
+    /// Returns the block number for this payload.
+    pub const fn block_number(&self) -> u64 {
+        self.as_v2().unwrap().payload_inner.block_number
+    }
+
+    /// Converts [`ExecutionPayloadV1`] to [`Block`].
+    ///
+    /// Caution: This does not set fields that are not part of the payload and only part of the
+    /// [`ExecutionPayloadSidecar`]:
```
emhane, line 62:
```suggestion
    /// [`OpExecutionPayloadSidecar`]:
```

emhane (changes requested):
nice so far! needs revised docs and also needs methods `v3` and `v3_mut`

crates/rpc-types-engine/src/sidecar.rs
mattsse, line 1: I don't think we need any of this

this just duplicates things we should already have on request

mattsse (changes requested):

crates/rpc-types-engine/src/sidecar.rs
varun-doshi: Can you please elaborate what changes I should make?

Cargo.toml
```diff
@@ -114,6 +114,7 @@ rstest = "0.24.0"
 ## crypto
 c-kzg = { version = "1.0", default-features = false }
 k256 = { version = "0.13", default-features = false, features = ["ecdsa"] }
+sha2 = { version = "0.10", default-features = false }
```
mattsse, line 117: this we don't need

mattsse (changes requested):
lets exlclude all the changes regarding requests

varun-doshi:
so should i also remove `try_into_block_with_sidecar`?
since that requires `requests_hash` which requires `sha256` and also `CancunPayloadFields` does not have a `RequestsOrHash` field similar to whats present in `alloy`

please correct me if im thinking about this wrong @mattsse

emhane:
> so should i also remove `try_into_block_with_sidecar`? since that requires `requests_hash` which requires `sha256` and also `CancunPayloadFields` does not have a `RequestsOrHash` field similar to whats present in `alloy`
> 
> please correct me if im thinking about this wrong @mattsse

we need `try_into_block_with_sidecar` for the parent beacon block root still

crates/rpc-types-engine/Cargo.toml
```diff
@@ -41,6 +41,9 @@ derive_more = { workspace = true, default-features = false, features = [
   "into",
 ]}
 
+#crypto
```
emhane, line 44:
```suggestion
# crypto
```
nitpick

crates/rpc-types-engine/src/sidecar.rs
```diff
@@ -87,6 +89,39 @@ impl OpExecutionPayloadSidecar {
...
+        }
+        Some(Requests::new(requests_vec))
+    }
+
+    /// Calculates or retrieves the requests hash.
+    ///
+    /// - If the `prague` field contains a list of requests, it calculates the requests hash
+    ///   dynamically.
+    /// - If it contains a precomputed hash (used for testing), it returns that hash directly.
+    pub fn requests_hash(&self) -> Option<B256> {
```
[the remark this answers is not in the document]
emhane: second this, this method should be removed since OP doesn't use EL requests

emhane (changes requested):
check out this review for guidance https://github.com/alloy-rs/op-alloy/pull/418, started impl this there but shrunk scope so as not to block op-reth development, for example a method for withdrawals root is missing

emhane:
blocked by https://github.com/alloy-rs/op-alloy/pull/435

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -19,3 +22,72 @@ pub enum OpExecutionPayload {
...
     V4(OpExecutionPayloadV4),
 }
+
+impl OpExecutionPayload {
+    /// Returns a reference to the V2 payload, if any.
+    pub const fn as_v2(&self) -> &ExecutionPayloadV2 {
+        match self {
+            Self::V2(payload) => payload,
+            Self::V3(payload) => &payload.payload_inner,
+            Self::V4(payload) => todo!(),
```
emhane, line 32:
```suggestion
            Self::V4(payload) => &payload.payload_inner.payload_inner,
```

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -19,3 +22,72 @@ pub enum OpExecutionPayload {
...
+    /// Caution: This does not set fields that are not part of the payload and only part of the
+    /// [`OpExecutionPayloadSidecar`]:
+    /// - parent_beacon_block_root
+    ///
+    /// See also: [`OpExecutionPayload::try_into_block_with_sidecar`]
+    pub fn try_into_block(self) -> Result<OpBlock, PayloadError> {
+        match self {
+            Self::V2(payload) => payload.try_into_block(),
+            Self::V3(payload) => payload.try_into_block(),
+            Self::V4(payload) => payload.payload_inner.try_into_block(),
```
emhane, line 70: when https://github.com/alloy-rs/op-alloy/pull/435 is merged we need
```suggestion
            Self::V4(payload) => payload.try_into_block(),
```

emhane (commented):
still v3 methods missing as mentioned in my last review

mattsse:
#435 is merged

I think this pr is almost done as well

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -19,3 +22,90 @@ pub enum OpExecutionPayload {
...
+    ///
+    /// See <https://github.com/ethereum/go-ethereum/blob/79a478bb6176425c2400e949890e668a3d9a3d05/core/beacon/types.go#L145>
+    pub fn try_into_block_with_sidecar<T: Decodable2718>(
+        self,
+        sidecar: &OpExecutionPayloadSidecar,
+    ) -> Result<Block<T>, PayloadError> {
+        let mut base_payload = self.try_into_block()?;
+        base_payload.header.parent_beacon_block_root = sidecar.parent_beacon_block_root();
+        base_payload.header.requests_hash = Some(EMPTY_ROOT_HASH);
+        base_payload.header.withdrawals_root = sidecar.withdrawals_root().copied();
```
emhane, line 107:
```suggestion
```
this shouldn't be in the sidecar, if this method is available on sidecar then this is a bug and it should be removed

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -19,3 +22,90 @@ pub enum OpExecutionPayload {
...
+    /// The log bloom is assumed to be validated during serialization.
+    ///
+    /// See <https://github.com/ethereum/go-ethereum/blob/79a478bb6176425c2400e949890e668a3d9a3d05/core/beacon/types.go#L145>
+    pub fn try_into_block_with_sidecar<T: Decodable2718>(
+        self,
+        sidecar: &OpExecutionPayloadSidecar,
+    ) -> Result<Block<T>, PayloadError> {
+        let mut base_payload = self.try_into_block()?;
+        base_payload.header.parent_beacon_block_root = sidecar.parent_beacon_block_root();
+        base_payload.header.requests_hash = Some(EMPTY_ROOT_HASH);
```
emhane, line 106: request hashes is none before pectra. I think the sidecar type will still need the Prague fields actually for this reason.

this method should check
```rust
base_payload.header.requests_hash = match sidecar.request_hashes() {
    Some(hashes) if !hashes.is_empty() => return Err(OpPayloadError::NonEmptyELRequests),
    Some(_) => Some(EMPTY_ROOT_HASH),
    None => None,
}
```
and same for blob versioned hashes. this needs a new error enum and everything. so let's remove this method from this pr and tackle it as sep issue.

emhane (changes requested):

crates/rpc-types-engine/src/payload/mod.rs
varun-doshi: removing this line `107`

crates/rpc-types-engine/src/payload/mod.rs
varun-doshi: removing the function `try_into_block_with_sidecar` to be tackled in separate PR

varun-doshi: I can take up the sub issue for sorting this function out when its created(unless you feel its a very advanced issue to tackle)

mattsse (approved):
lgtm, pending @emhane

crates/rpc-types-engine/src/payload/mod.rs
emhane: yes pls do

emhane (approved):

Issue #425, opened 2025-02-10
[Feature] custom deserde impl for Opexecutionpayload


mattsse:
### Component

rpc-types

### Describe the feature you would like

as described in https://github.com/alloy-rs/alloy/issues/2004

[alloy/crates/rpc-types-engine/src/payload.rs](https://github.com/alloy-rs/alloy/blob/0fbc37da5178681e51e93d5274681f85c874702a/crates/rpc-types-engine/src/payload.rs#L827-L832)

Lines 827 to 832 in [0fbc37d](https://github.com/alloy-rs/alloy/commit/0fbc37da5178681e51e93d5274681f85c874702a)

```
 /// An execution payload, which can be either [ExecutionPayloadV1], [ExecutionPayloadV2], or 
 /// [ExecutionPayloadV3]. 
 #[derive(Clone, Debug, PartialEq, Eq)] 
 #[cfg_attr(feature = "serde", derive(serde::Serialize))] 
 #[cfg_attr(feature = "serde", serde(untagged))] 
 pub enum ExecutionPayload { 
right now if we have a ExecutionPayloadV3 json, we can still do:
let payload: ExecutionPayloadV2 = serde_json::from_str(json).unwrap();
```

and extra fields just implicitly get ignored. but obv you can no longer roundtrip then.

see also https://github.com/alloy-rs/alloy/pull/2017

see also https://github.com/alloy-rs/alloy/issues/2015

## TODO 

* replicate https://github.com/alloy-rs/alloy/pull/2017 for OpExecutionpayload

varun-doshi:
I can take this one

emhane:
closed in https://github.com/alloy-rs/op-alloy/pull/436

PR #436, opened 2025-02-13
custom deserialize impl for OpExecutionPayload

varun-doshi:
## Motivation

#425 

## Solution

Create custom `deserialise` impl for `OpExecutionPayload`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


varun-doshi:
I expect some mistakes in this PR to be found on review...will correct as advised

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -23,6 +25,245 @@ pub enum OpExecutionPayload {
...
+                    GasLimit,
+                    GasUsed,
+                    Timestamp,
+                    ExtraData,
+                    BaseFeePerGas,
+                    BlockHash,
+                    Transactions,
+                    Withdrawals,
+                    BlobGasUsed,
+                    ExcessBlobGas,
```
mattsse, line 66: for v4 we also need withdrawalsroot

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -23,6 +25,245 @@ pub enum OpExecutionPayload {
...
+                    block_hash: block_hash
+                        .ok_or_else(|| serde::de::Error::missing_field("blockHash"))?,
+                    transactions: transactions
+                        .ok_or_else(|| serde::de::Error::missing_field("transactions"))?,
+                };
+
+                if let (Some(blob_gas_used), Some(excess_blob_gas)) =
+                    (blob_gas_used, excess_blob_gas)
+                {
+                    return Ok(OpExecutionPayload::V3(ExecutionPayloadV3 {
```
mattsse, line 221: here we need to check if withdrawalsroot is set, if so then this v4

mattsse (changes requested):
nice,

can we also duplicate the test cases from the referenced alloy pr here and add one for v4?

varun-doshi:
Note: Added a `as_v4()` method since it was required to test the deserde
Can remove if its an issue
@mattsse

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -23,6 +25,269 @@ pub enum OpExecutionPayload {
...
+                                excess_blob_gas,
+                            },
+                            withdrawals_root,
+                        }));
+                    };
+                // return V3 is withdrawals_root is absent, but blob_gas_used and excess_blob_gas is present
+                } else if let (Some(blob_gas_used), Some(excess_blob_gas)) =
+                    (blob_gas_used, excess_blob_gas)
+                {
+                    return Ok(OpExecutionPayload::V3(ExecutionPayloadV3 {
```
mattsse, line 243: I think this can be simplified because we create the v3 here twice now and have the some blob gas checks,

we can do those first and only then decide if we need v4 or v3 based on the withdrawalsroot

crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -23,6 +25,269 @@ pub enum OpExecutionPayload {
...
+                        payload_inner: ExecutionPayloadV2 {
+                            payload_inner: v1,
+                            withdrawals: withdrawals.unwrap(),
```
mattsse, line 246: why is this unwrap safe?

mattsse (changes requested):
a few more nits, and needs a few fixes for no-std, see CI

mattsse (approved):
ty

Issue #439, opened 2025-02-14
[Feature] Impl conversion to block with sidecar for `OpExecutionPayload`


emhane:
### Component

rpc-types-engine

### Describe the feature you would like

Impl `OpExecutionPayload::try_into_block_with_sidecar`

emhane:
@varun-doshi heads up I'm doing this now just because I need to unblock myself on op-reth

PR #441, opened 2025-02-14
feat(l2-withdrawals): impl conversion payload + sidecar into block 

emhane:
Closes https://github.com/alloy-rs/op-alloy/issues/439

- Adds `OpPayloadError` as extension of `PayloadError` with op variants


mattsse (approved):
lgtm

Issue #449, opened 2025-02-22
[Feature] impl TryFrom AnyRpcTransaction for OpTxEnvelope


mattsse:
### Component

op-alloy

### Describe the feature you would like

we need a few additional conversion helpers for OpTxEnvelope

ref 

https://github.com/paradigmxyz/reth/blob/d39e24d9f363334647193c142bd980cfa6b061ab/crates/optimism/primitives/src/alloy_compat.rs#L10-L10

which we need for OpTxEnvelope, we already have:

https://github.com/alloy-rs/op-alloy/blob/856e41042527a6ba4e0e733980435c3aa6192135/crates/consensus/src/transaction/envelope.rs#L375-L378

## TODO 

* impl TryFrom AnyRpcTransaction for OpTxEnvelope in alloy_compat.rs

also needs test case:

https://github.com/paradigmxyz/reth/blob/d39e24d9f363334647193c142bd980cfa6b061ab/crates/optimism/primitives/src/alloy_compat.rs#L66-L66

FYI @varun-doshi

I believe we're likely missing a few more but I'll find out

varun-doshi:
Understood
I can take this up

PR #454, opened 2025-02-23
feat: impl AnyRpcTransaction for OpTxEnvelope

varun-doshi:
## Motivation

Ref #449 

## Solution
Implement TryFrom<AnyRpcTransaction> for OpTxEnvelope 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/alloy_compat.rs
+++ b/crates/consensus/src/alloy_compat.rs
@@ -122,6 +122,7 @@ mod tests {
   "value": "0x0"
 }"#;
         let tx: AnyRpcTransaction = serde_json::from_str(json).unwrap();
-        OpTxEnvelope::try_from(tx).unwrap();
+        let tx = OpTxEnvelope::try_from(tx).unwrap();
+        assert!(tx.is_deposit());
     }
 }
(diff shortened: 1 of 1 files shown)


crates/consensus/src/alloy_compat.rs
```diff
@@ -36,6 +38,46 @@ impl TryFrom<AnyTxEnvelope> for OpTxEnvelope {
...
+                Ok(OpTxEnvelope::Eip2930(tx.try_into().map_err(|_| {
+                    ConversionError::Custom("Failed to convert Eip2930 Tx".to_string())
+                })?))
+            }
+            AnyTxEnvelope::Ethereum(TxEnvelope::Eip1559(tx)) => {
+                Ok(OpTxEnvelope::Eip1559(tx.try_into().map_err(|_| {
+                    ConversionError::Custom("Failed to convert Eip1559 Tx".to_string())
+                })?))
+            }
+            AnyTxEnvelope::Ethereum(TxEnvelope::Eip7702(tx)) => {
```
mattsse, line 63: I think this can be simplified now that we have a tryinto for TxEnvelope:

https://github.com/alloy-rs/op-alloy/blob/0c6d6dbb2da0a13f81643759046410190e2c59e4/crates/consensus/src/transaction/envelope.rs#L361-L361

mattsse (changes requested):
great, but we can simplify this a bit by reusing existing conversions

mattsse (approved):

PR #455, opened 2025-02-26
feat(l2-withdrawals-root): Add conversions for `OpExecutionData`

emhane:
- Adds missing variant `OpExecutionPayload::V1`, needed to support conversion from `ExecutionPayloadInputV2`
- Fixes buggy `as_mut` methods for `OpExecutionPayload`
- Adds conversion from `newPayload` args to `OpExecutionData`

### DIFF
--- a/crates/rpc-types-engine/src/payload/mod.rs
+++ b/crates/rpc-types-engine/src/payload/mod.rs
@@ -508,7 +508,7 @@ mod tests {
         assert_eq!(serde_json::to_string(&payload).unwrap(), response_v2);
 
         let payload_v2: ExecutionPayloadV2 = serde_json::from_str(response_v2).unwrap();
-        assert_eq!(payload.as_v2(), &payload_v2);
+        assert_eq!(payload.as_v2(), Some(&payload_v2));
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -17,6 +20,8 @@ use error::OpPayloadError;
 #[cfg_attr(feature = "serde", derive(serde::Serialize))]
 #[cfg_attr(feature = "serde", serde(untagged))]
 pub enum OpExecutionPayload {
+    /// V1 payload
+    V1(ExecutionPayloadV1),
```
mattsse, line 24: why do we need this again, I thought there is no v1?

mattsse (commented):
question about v1

mattsse (approved):
> needed to support conversion from ExecutionPayloadInputV2

ah I should have read more closely, that makes sense because this type is used as rpc arg so we need to handle this somehow

crates/rpc-types-engine/src/payload/mod.rs
emhane: the disjoint versioning of: new payload method, engine payload AND engine payload body...are seriously confusing

PR #456, opened 2025-02-26
Remove deprecated key lint from manifests

emhane:
Fixes "unused manifest key `lints`" error: https://github.com/alloy-rs/op-alloy/actions/runs/13541694801/job/37844009342?pr=455

### DIFF
--- a/crates/consensus/Cargo.toml
+++ b/crates/consensus/Cargo.toml
@@ -11,9 +11,6 @@ homepage.workspace = true
 repository.workspace = true
 exclude.workspace = true
 
-[lints]
-workspace = true
-
 [dependencies]
 # Alloy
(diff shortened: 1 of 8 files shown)


mattsse (approved):

mattsse:
I think this an edition2024 thing, so can also hold off here

emhane:
> I think this an edition2024 thing, so can also hold off here

yes and need to double check if this fix is actually needed, seems like maybe root cause of this deny check error was edition mismatch

refcell:
I think this lint inheritance is still respected by cargo, but maybe we don't need anymore?

emhane:
think this error was funnily triggered by underlying issue

PR #457, opened 2025-02-26
Bump msrv to 1.85

emhane:
Bumps MSRV to 1.85


clabby:
> fine with me but iirc @clabby flagged some msrv issues on kona coming from other teams, so we should hold off here until we checked

Yeah, we need to wait on this unfortunately. The RiscZero zkVM toolchain is getting to `1.83` by EOW, and SP-1's is on `1.82` iirc.

refcell (approved):
Let's get this in now since kona's been bumped a while ago.

PR #460, opened 2025-02-27
Add conversions from block to payload

emhane:
Adds conversion from block to `OpExecutionPayload` and `OpExecutionData`

### DIFF
--- a/crates/rpc-types-engine/src/payload/mod.rs
+++ b/crates/rpc-types-engine/src/payload/mod.rs
@@ -5,9 +5,9 @@ pub mod v3;
 pub mod v4;
 
 use crate::{OpExecutionPayloadSidecar, OpExecutionPayloadV4};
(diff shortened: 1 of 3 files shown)


crates/rpc-types-engine/src/payload/mod.rs
```diff
@@ -291,6 +291,60 @@ impl<'de> serde::Deserialize<'de> for OpExecutionPayload {
...
+    ) -> (Self, OpExecutionPayloadSidecar)
+    where
+        T: Encodable2718 + Transaction,
+        H: BlockHeader,
+    {
+        let sidecar = OpExecutionPayloadSidecar::from_block(block);
+
+        let execution_payload = match block.withdrawals_root() {
+            Some(withdrawals_root) if sidecar.isthmus().is_some() => {
+                // block with (empty) request hashes: V4
```
mattsse, line 325: this works

mattsse (approved):

PR #461, opened 2025-02-27
Remove redundant method for v4 payload

emhane:
No need for method `OpExecutionPayloadV4::withdrawals` and the types it imports, since the inner payload is in a publicly accessible field anyways. This list of withdrawals should always be empty.


mattsse (commented):
why remove this?

having withdrawals as fn is a lot nicer

mattsse (approved):

PR #467, opened 2025-03-07
feat(consensus): Into Recovered

refcell:
### Description

Adds support to `OpTxEnvelope` for `Into<Recovered>`

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -347,6 +347,32 @@ impl OpTxEnvelope {
         }
     }
 
+    /// Recover the signer of the transaction.
+    #[cfg(feature = "k256")]
+    pub fn recover_signer(
+        &self,
+    ) -> Result<alloy_primitives::Address, alloy_primitives::SignatureError> {
+        match self {
+            Self::Legacy(tx) => tx.recover_signer(),
+            Self::Eip2930(tx) => tx.recover_signer(),
+            Self::Eip1559(tx) => tx.recover_signer(),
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

refcell:
I'm not sure we want the deposit tx to error here. Should we instead use address::zero so it succeeds in recovery? Less familiar with how the `alloy_consensus::transaction::Recovered` type is used.

refcell:
Turns out we don't need this for kona, can just do a `alloy_consensus::transaction::Recovered::new_unchecked(OpTxEnveleope::TxDeposit { ... }, Address::ZERO)` in the tests. Maybe best not to introduce so it isn't erroneously used?

mattsse:
> Should we instead use address::zero

the way we do this in reth is, we use the from address https://github.com/paradigmxyz/reth/blob/c3c152f62c70378f6008728df84cc1334ed14fbc/crates/optimism/primitives/src/transaction/signed.rs#L107-L111

emhane (changes requested):
I agree with @refcell that it's dangerous to implement it this way for op txns. any batch processing of transactions in op-reth code that returns early on error, will return early if a deposit tx is in the batch. this could lead to many hard to discover bugs. it doesn't make sense for shared reth/op-reth code to match on some error variant related to deposit tx to filter out this error and not return early.

I think it makes our lives easier if we add method `fn as_signable(&self) -> Option<T>` and get used to having to map any calls related to signer. I think all the non-deposit txns implement `SignableTransaction` and `SignerRecoverable`, so we can even add those trait bounds to `T`.

PR #471, opened 2025-03-13
feat(rpc-types-engine): Support V4 network payload

clabby:
## Overview

Adds support for the V4 CL network payload, introduced with Isthmus.


refcell (approved):
Nice thanks for getting this in

PR #472, opened 2025-03-14
fix(l2-withdrawals-root): `OpExecutionPayloadEnvelopeV4` missing v4 payload

emhane:
Bug: `OpExecutionPayloadEnvelopeV4` contains v3 not op v4 payload
Fix: change to op v4 payload


mattsse (approved):
ah yeah -.-

PR #477, opened 2025-03-25
feat: move safety level

mattsse:
ref https://github.com/paradigmxyz/reth/pull/15105

fyi @SozinM


SozinM:
wait, i already moved everything in my PR

mattsse:
there's no reason to define these twice, this can just go in here

SozinM:
my idea is to keep all kona related deps in one place and then remove them once kona is good to be used in reth
and as it only use with supervisor i just put them in primitives under supervisor feature

PR #478, opened 2025-03-25
feat: add input mut

mattsse:
blocked by https://github.com/alloy-rs/core/pull/921

same as https://github.com/alloy-rs/alloy/pull/2244

for https://github.com/paradigmxyz/reth/issues/15245


emhane (approved):
feels very counterintuitive to have input mut and break functional programming paradigm just for otterscan, but ok

PR #482, opened 2025-03-28
fix(rpc-types-engine): Fix `PayloadHash` for V3 + V4 topic

clabby:
## Overview

Fixes the payload hash for the V3 + V4 topic.

spec: https://specs.optimism.io/protocol/rollup-node-p2p.html#block-signatures


refcell (approved):

mattsse (approved):

clabby:
Pulling into draft for a sec; Want to actually check that we validate block signatures on ingress w/ this first.

clabby:
Yeah this works. Merging.

PR #483, opened 2025-03-29
feat: add hash ref function

mattsse:
hash fn is a bit of a mess rn, we have hash, hash_ref and tx_hash.

this follows the convention that hash should return a ref.

before next breaking release we should make this more consistent

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -448,14 +448,19 @@ impl OpTxEnvelope {
     }
 
     /// Returns the inner transaction hash.
-    pub fn tx_hash(&self) -> B256 {
+    pub fn hash(&self) -> &B256 {
         match self {
-            Self::Legacy(tx) => *tx.hash(),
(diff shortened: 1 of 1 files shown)


crates/consensus/src/transaction/envelope.rs
```diff
@@ -448,16 +448,21 @@ impl OpTxEnvelope {
...
+            Self::Legacy(tx) => tx.hash(),
+            Self::Eip1559(tx) => tx.hash(),
+            Self::Eip2930(tx) => tx.hash(),
+            Self::Eip7702(tx) => tx.hash(),
+            Self::Deposit(tx) => tx.hash_ref(),
         }
     }
 
+    /// Returns the inner transaction hash.
+    pub fn tx_hash(&self) -> B256 {
```
emhane, line 462: is more straight forward if this method is called `hash_copied`, alt `hash` and call the other one `hash_ref`

emhane (approved):
i'm for having a method that returns a reference to the 32 bytes, agree that the naming needs revision

PR #485, opened 2025-04-01
feat(rpc-types-engine): SSZ Payload Encoding

refcell:
### Description

Implements ssz payload encoding for gossiping `OpNetworkPayloadEnvelope`s via consensus p2p.

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -448,19 +448,19 @@ impl OpTxEnvelope {
     }
 
     /// Returns the inner transaction hash.
-    pub fn hash(&self) -> &B256 {
+    pub fn hash(&self) -> B256 {
         match self {
-            Self::Legacy(tx) => tx.hash(),
-            Self::Eip1559(tx) => tx.hash(),
-            Self::Eip2930(tx) => tx.hash(),
-            Self::Eip7702(tx) => tx.hash(),
-            Self::Deposit(tx) => tx.hash_ref(),
+            Self::Legacy(tx) => *tx.hash(),
+            Self::Eip1559(tx) => *tx.hash(),
+            Self::Eip2930(tx) => *tx.hash(),
+            Self::Eip7702(tx) => *tx.hash(),
+            Self::Deposit(tx) => tx.hash(),
         }
     }
 
     /// Returns the inner transaction hash.
     pub fn tx_hash(&self) -> B256 {
-        *self.hash()
+        self.hash()
     }
 
     /// Return the length of the inner txn, including type byte length
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -448,19 +448,19 @@ impl OpTxEnvelope {
     }
 
     /// Returns the inner transaction hash.
-    pub fn hash(&self) -> B256 {
+    pub fn hash(&self) -> &B256 {
         match self {
-            Self::Legacy(tx) => *tx.hash(),
-            Self::Eip1559(tx) => *tx.hash(),
-            Self::Eip2930(tx) => *tx.hash(),
-            Self::Eip7702(tx) => *tx.hash(),
-            Self::Deposit(tx) => tx.hash(),
+            Self::Legacy(tx) => tx.hash(),
+            Self::Eip1559(tx) => tx.hash(),
+            Self::Eip2930(tx) => tx.hash(),
+            Self::Eip7702(tx) => tx.hash(),
+            Self::Deposit(tx) => tx.hash_ref(),
         }
     }
 
     /// Returns the inner transaction hash.
     pub fn tx_hash(&self) -> B256 {
-        self.hash()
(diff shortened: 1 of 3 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #485 👈
* `main`

crates/consensus/src/transaction/envelope.rs
```diff
@@ -448,19 +448,19 @@ impl OpTxEnvelope {
     }
 
     /// Returns the inner transaction hash.
-    pub fn hash(&self) -> &B256 {
+    pub fn hash(&self) -> B256 {
```
mattsse, line 451: I need this to return a reference.

hash fns need a cleanup, but this change doesn't belong in this pr

crates/rpc-types-engine/src/envelope.rs
```diff
@@ -237,6 +237,27 @@ impl OpNetworkPayloadEnvelope {
...
 
+    /// Encodes a payload envelope as a snappy-compressed byte array.
+    #[cfg(feature = "std")]
+    pub fn encode_v3(&self) -> Result<Vec<u8>, PayloadEnvelopeEncodeError> {
+        let execution_payload_v3 = match &self.payload {
+            OpExecutionPayload::V3(execution_payload_v3) => execution_payload_v3,
+            _ => return Err(PayloadEnvelopeEncodeError::WrongVersion),
+        };
+
+        use ssz::Encode;
```
mattsse, line 248: let's move this top of function

crates/rpc-types-engine/src/envelope.rs
```diff
@@ -270,6 +291,17 @@ impl OpNetworkPayloadEnvelope {
...
+/// Errors that can occur when encoding a payload envelope.
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub enum PayloadEnvelopeEncodeError {
+    /// Wrong versions of the payload.
+    #[error("Wrong version of the payload")]
+    WrongVersion,
+    /// An error occured during snap encoding.
+    #[error(transparent)]
+    SnapEncoding(#[from] snap::Error),
+}
```
mattsse, line 303: doe we need this additional error if we already have

PayloadEnvelopeError

crates/consensus/src/transaction/envelope.rs
refcell: Hrm ci was failing

refcell: Because `TxDeposit` hash func

crates/rpc-types-engine/src/envelope.rs
refcell: Are you alright if I add a variant to that to say the snap encoding failed?

refcell: As well as the wrong version actually too

refcell: Basically - just need these variants in `PayloadEnvelopeError` to be able to use it

crates/rpc-types-engine/src/envelope.rs
```diff
@@ -203,6 +229,32 @@ impl OpNetworkPayloadEnvelope {
...
+
+        use ssz::Encode;
+        let mut data = Vec::new();
+        let mut sig_bytes = self.signature.as_bytes();
+        // Normalize parity bit
+        if self.signature.v() {
+            sig_bytes[64] = 0x01;
+        } else {
+            sig_bytes[64] = 0x00;
+        }
```
clabby, line 248: Small nit - let's use `Parity::y_parity_byte` here & in the other parity bit normalizations: https://docs.rs/alloy-primitives/latest/alloy_primitives/enum.Parity.html#method.y_parity_byte

```suggestion
        sig_bytes[64] = self.signature.v().y_parity_byte();
```

refcell: Oh nice!

clabby (approved):
Looks great, I'm ambivalent on the new error type - whatever y'all decide on is cool with me here.

crates/rpc-types-engine/src/envelope.rs
```diff
@@ -268,6 +326,38 @@ impl OpNetworkPayloadEnvelope {
...
+    pub fn encode_v4(&self) -> Result<Vec<u8>, PayloadEnvelopeEncodeError> {
+        use ssz::Encode;
+        let execution_payload_v4 = match &self.payload {
+            OpExecutionPayload::V4(execution_payload_v4) => execution_payload_v4,
+            _ => return Err(PayloadEnvelopeEncodeError::WrongVersion),
+        };
+
+        let mut data = Vec::new();
+        let mut sig = self.signature.as_bytes();
+        sig[64] = self.signature.v() as u8;
```
clabby, line 341: Ah nice. I didn't notice we aliased the import of the deprecated `Signature` type -> `PrimitiveSignature` in my prev comment. This works too with the new type 😄

mattsse (approved):

PR #486, opened 2025-04-02
chore: implement serde_bincode_compat for envelope

shane-moore:
- This PR also helps close out https://github.com/paradigmxyz/reth/issues/15377
- Related implementation: https://github.com/alloy-rs/alloy/pull/2263
- Running test via `cargo test --features "serde serde-bincode-compat arbitrary k256" test_op_tx_envelope_bincode_roundtrip_arbitrary` shows successful test. Please lmk if it's robust enough or if there's some pointers to make this implementation more extensive 🤙 
- Note that you'll need `alloy-consensus = { workspace = true, features = ["serde", "serde-bincode-compat" ] }` and `[features]
default = ["std", "serde", "serde-bincode-compat", "arbitrary"]` in the `Cargo.toml`

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -1,14 +1,14 @@
 use crate::{OpTxType, OpTypedTransaction, TxDeposit};
 use alloy_consensus::{
-    Sealable, Sealed, Signed, Transaction, TxEip1559, TxEip2930, TxEip7702, TxEnvelope, TxLegacy,
-    Typed2718, transaction::RlpEcdsaDecodableTx,
+    transaction::RlpEcdsaDecodableTx, Sealable, Sealed, Signed, Transaction, TxEip1559, TxEip2930,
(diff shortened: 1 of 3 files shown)


crates/consensus/src/transaction/envelope.rs
```diff
@@ -629,12 +629,173 @@ mod serde_from {
...
+            /// Borrowed EIP-2930 transaction data.
+            #[serde(borrow)]
+            transaction: TxEip2930<'a>,
+        },
+        /// EIP-1559 variant.
+        Eip1559 {
+            /// Transaction signature.
+            signature: Signature,
+            /// Borrowed EIP-1559 transaction data.
+            #[serde(borrow)]
```
mattsse, line 668: ah nice, this way we dont neven need a cow, never thought about this and now I feel stupid :)

mattsse (approved):
great!

crates/consensus/src/transaction/envelope.rs
shane-moore: Super cool, right!

PR #489, opened 2025-04-04
feat: added recover helpers for OpPayloadAttributes

### LINKED ISSUE #488: [Feature] Add recover helpers for OpPayloadAttributes
### Component

op-alloy

for 

https://github.com/alloy-rs/op-alloy/blob/27a0bc50de7af04bb0d487b6342f18c15c49dcb3/crates/rpc-types-engine/src/attributes.rs#L13-L19

we want a few functions that
* decode the transactions (OpTxEnvelope)
* and recover the senders: Recovered
* and return them as `WithEncoded`

## TODO
* add helper fns that return iterators for `Recovered<OpTxEnvelope>` and `WithEncoded<OpTxEnvelope>` and `WithEncoded<Recovered<OpTxEnvelope>>`

maybe @rimeeeeee 

### Additional context

_No response_

Rimeeeeee:
fixes #488

### DIFF
--- a/crates/rpc-types-engine/src/attributes.rs
+++ b/crates/rpc-types-engine/src/attributes.rs
@@ -4,6 +4,7 @@ use alloc::vec::Vec;
 use alloy_consensus::transaction::Recovered;
 use alloy_eips::{Decodable2718, eip1559::BaseFeeParams, eip2718::WithEncoded};
 use alloy_primitives::{B64, Bytes, SignatureError};
+use alloy_rlp::Result;
 use alloy_rpc_types_engine::PayloadAttributes;
 use op_alloy_consensus::{
     EIP1559ParamError, OpTxEnvelope, decode_eip_1559_params, encode_holocene_extra_data,
@@ -91,26 +92,20 @@ impl OpPayloadAttributes {
         })
     }
 
-    /// Helper: returns first successfully recovered transaction, if any.
-    pub fn try_into_recovered(&self) -> Option<Recovered<OpTxEnvelope>> {
-        self.try_to_recovered().find_map(Result::ok)
-    }
-
     /// Returns iterator over decoded transactions with their original encoded bytes.
     pub fn decoded_transactions_with_encoded(
         &self,
     ) -> impl Iterator<Item = Result<WithEncoded<OpTxEnvelope>, SignatureError>> + '_ {
-        self.transactions.iter().flatten().map(|tx_bytes| {
-            let env = OpTxEnvelope::decode_2718(&mut tx_bytes.as_ref())
-                .map_err(|_| SignatureError::FromBytes("Failed to decode OpTxEnvelope"))?;
+        let mut decoded_iter = self.try_to_decoded();
 
-            let op_tx = OpTxEnvelope::try_from(env)
-                .map_err(|_| SignatureError::FromBytes("Failed to convert envelope"))?;
+        self.transactions.iter().flatten().map(move |tx_bytes| {
+            let decoded = decoded_iter
+                .next()
+                .ok_or(SignatureError::FromBytes("Mismatched decoded transaction"))?;
 
-            Ok(WithEncoded::new(tx_bytes.clone(), op_tx))
+            decoded.map(|op_tx| WithEncoded::new(tx_bytes.clone(), op_tx))
         })
     }
-
     /// Returns iterator over successfully recovered transactions with encoded bytes.
     pub fn recovered_transactions_with_encoded(
         &self,
--- a/crates/rpc-types-engine/src/attributes.rs
+++ b/crates/rpc-types-engine/src/attributes.rs
@@ -1,10 +1,13 @@
 //! Optimism-specific payload attributes.
 
 use alloc::vec::Vec;
-use alloy_eips::eip1559::BaseFeeParams;
+use alloy_consensus::transaction::Recovered;
+use alloy_eips::{Decodable2718, eip1559::BaseFeeParams, eip2718::WithEncoded};
 use alloy_primitives::{B64, Bytes};
 use alloy_rpc_types_engine::PayloadAttributes;
-use op_alloy_consensus::{EIP1559ParamError, decode_eip_1559_params, encode_holocene_extra_data};
+use op_alloy_consensus::{
+    EIP1559ParamError, OpTxEnvelope, decode_eip_1559_params, encode_holocene_extra_data,
+};
 
 /// Optimism Payload Attributes
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
@@ -51,6 +54,41 @@ impl OpPayloadAttributes {
     /// Returns (`elasticity`, `denominator`)
     pub fn decode_eip_1559_params(&self) -> Option<(u32, u32)> {
         self.eip_1559_params.map(decode_eip_1559_params)
+    }
+    /// Returns an iterator over `WithEncoded<OpTxEnvelope>`
+    pub fn decoded_transactions_with_encoded(
+        &self,
+    ) -> impl Iterator<Item = WithEncoded<OpTxEnvelope>> + '_ {
+        self.transactions.iter().flatten().filter_map(|tx_bytes| {
+            OpTxEnvelope::decode_2718(&mut tx_bytes.as_ref())
+                .ok()
+                .and_then(|env| OpTxEnvelope::try_from(env).ok())
+                .map(|env| WithEncoded::new(tx_bytes.clone(), env))
+        })
+    }
+    /// Returns an iterator over `WithEncoded<Recovered<OpTxEnvelope>>`
+    pub fn recovered_transactions_with_encoded(
+        &self,
+    ) -> impl Iterator<Item = WithEncoded<Recovered<OpTxEnvelope>>> + '_ {
+        self.transactions
+            .iter()
+            .flatten()
+            .zip(self.recovered_transactions())
+            .map(|(tx_bytes, recovered)| WithEncoded::new(tx_bytes.clone(), recovered))
+    }
+    /// Returns an iterator over `Recovered<OpTxEnvelope>`
+    pub fn recovered_transactions(&self) -> impl Iterator<Item = Recovered<OpTxEnvelope>> + '_ {
+        self.transactions.iter().flatten().filter_map(|tx_bytes| {
+            let env = OpTxEnvelope::decode_2718(&mut tx_bytes.as_ref()).ok()?;
+            let txenv = env.try_into_eth_envelope().ok()?;
+            let recovered = txenv.try_into_recovered().ok()?;
+
+            // Convert EthereumTxEnvelope back to OpTxEnvelope
+            match OpTxEnvelope::try_from_eth_envelope(recovered.inner().clone()) {
+                Ok(op_tx) => Some(Recovered::new_unchecked(op_tx, recovered.signer().clone())),
+                Err(_) => None,
+            }
+        })
     }
 }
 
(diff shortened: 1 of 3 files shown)


Rimeeeeee:
@mattsse ,is the approach okay?

crates/rpc-types-engine/src/attributes.rs
```diff
@@ -52,6 +55,46 @@ impl OpPayloadAttributes {
...
+            let env = OpTxEnvelope::decode_2718(&mut tx_bytes.as_ref()).ok()?;
+            let txenv = env.try_into_eth_envelope().ok()?;
+            let recovered = txenv.try_into_recovered().ok()?;
+
+            // Convert EthereumTxEnvelope back to OpTxEnvelope
+            match OpTxEnvelope::try_from_eth_envelope(recovered.inner().clone()) {
+                Ok(op_tx) => Some(Recovered::new_unchecked(op_tx, recovered.signer().clone())),
+                Err(_) => None,
+            }
+        })
```
mattsse, line 99: what we can do here is reuse the other function and then simply remove the outer `WithEncoded` layer, then we don't need to add the code twice

crates/rpc-types-engine/src/attributes.rs
```diff
@@ -70,11 +70,16 @@ impl OpPayloadAttributes {
...
     ) -> impl Iterator<Item = WithEncoded<Recovered<OpTxEnvelope>>> + '_ {
-        self.transactions
-            .iter()
-            .flatten()
-            .zip(self.recovered_transactions())
-            .map(|(tx_bytes, recovered)| WithEncoded::new(tx_bytes.clone(), recovered))
+        self.transactions.iter().flatten().filter_map(|tx_bytes| {
+            let env = OpTxEnvelope::decode_2718(&mut tx_bytes.as_ref()).ok()?;
+            let txenv = env.try_into_eth_envelope().ok()?;
+            let recovered = txenv.try_into_recovered().ok()?;
```
emhane, line 76: this code can be moved into a separate function `pub fn try_into_recovered(bytes: &[u8]) -> Result<RecoveredType, ErrorType>` and called here and in the body of the method below `recovered_transactions`

emhane (changes requested):

crates/rpc-types-engine/src/attributes.rs
```diff
@@ -52,6 +55,46 @@ impl OpPayloadAttributes {
...
+            // Convert EthereumTxEnvelope back to OpTxEnvelope
+            match OpTxEnvelope::try_from_eth_envelope(recovered.inner().clone()) {
+                Ok(op_tx) => Some(WithEncoded::new(tx_bytes.clone(), Recovered::new_unchecked(op_tx, recovered.signer().clone()))),
+                Err(_) => None,
+            }
+        })
```
mattsse: ideally we can reuse a single function that does the encoding and then simply layer the things we need on top, like sender recovery

crates/rpc-types-engine/src/attributes.rs
```diff
@@ -52,6 +55,46 @@ impl OpPayloadAttributes {
...
+    pub fn decoded_transactions_with_encoded(
+        &self,
+    ) -> impl Iterator<Item = WithEncoded<OpTxEnvelope>> + '_ {
+        self.transactions.iter().flatten().filter_map(|tx_bytes| {
+            OpTxEnvelope::decode_2718(&mut tx_bytes.as_ref())
+                .ok()
+                .and_then(|env| OpTxEnvelope::try_from(env).ok())
+                .map(|env| WithEncoded::new(tx_bytes.clone(), env))
+        })
+    }
```
mattsse, line 68: oh actually now that I see that this is fallibe, we should return an iterator over `Resul<WithEncoded`

which can be collected into `Result<Vec` by the caller

mattsse (changes requested):
great start, a few more nits.

let's try to reuse as much code as possible by combining these iter functions

for example, first step is always decoding, then next step would be a map etc..

crates/rpc-types-engine/src/attributes.rs
Rimeeeeee: @mattsse do we have specific `Error` while working with `Res<WithEncoded `and `Result<Vec` ?

Rimeeeeee:
@mattsse tried to address it, pls check it

crates/rpc-types-engine/src/attributes.rs
```diff
@@ -52,6 +55,48 @@ impl OpPayloadAttributes {
...
+
+    /// Returns an iterator over `Result<WithEncoded<OpTxEnvelope>>`
+    pub fn decoded_transactions_with_encoded(
+        &self,
+    ) -> impl Iterator<Item = Result<WithEncoded<OpTxEnvelope>, Box<dyn std::error::Error>>> + '_
+    {
+        self.transactions.iter().flatten().map(|tx_bytes| {
+            let env = OpTxEnvelope::decode_2718(&mut tx_bytes.as_ref())
+                .map_err(|e| Box::<dyn std::error::Error>::from(format!("Decode error: {}", e)))?;
```
emhane, line 77:
```suggestion
                .map_err(|e| Box::<dyn std::error::Error>::from(format!("Decode error: {e}")))?;
```
nitpick

crates/rpc-types-engine/src/attributes.rs
```diff
@@ -52,6 +55,48 @@ impl OpPayloadAttributes {
...
+
+            let op_tx = OpTxEnvelope::try_from(env).map_err(|e| {
+                Box::<dyn std::error::Error>::from(format!("Conversion error: {}", e))
```
emhane, line 80:
```suggestion
                Box::<dyn std::error::Error>::from(format!("Conversion error: {e}"))
```
nitpick

crates/rpc-types-engine/src/attributes.rs
```diff
@@ -52,6 +55,48 @@ impl OpPayloadAttributes {
...
+
+            Ok(WithEncoded::new(tx_bytes.clone(), op_tx))
+        })
+    }
+    /// Returns an iterator over `WithEncoded<Recovered<OpTxEnvelope>>`
+    pub fn recovered_transactions_with_encoded(
+        &self,
+    ) -> impl Iterator<Item = WithEncoded<Recovered<OpTxEnvelope>>> + '_ {
+        self.transactions.iter().flatten().filter_map(|tx_bytes| {
+            let recovered = self.try_into_recovered(tx_bytes)?;
```
emhane, line 91: do we want to trace log that don't decode?

crates/rpc-types-engine/src/attributes.rs
```diff
@@ -52,6 +55,48 @@ impl OpPayloadAttributes {
...
+                .map_err(|e| Box::<dyn std::error::Error>::from(format!("Decode error: {}", e)))?;
+
+            let op_tx = OpTxEnvelope::try_from(env).map_err(|e| {
+                Box::<dyn std::error::Error>::from(format!("Conversion error: {}", e))
+            })?;
+
+            Ok(WithEncoded::new(tx_bytes.clone(), op_tx))
+        })
+    }
+    /// Returns an iterator over `WithEncoded<Recovered<OpTxEnvelope>>`
```
emhane, line 86: docs should mention that those that don't decode are filtered out

crates/rpc-types-engine/src/attributes.rs
Rimeeeeee: addressed the changes :)

Rimeeeeee:
@mattsse @emhane lmk if i have to do something else :)

mattsse (changes requested):
we need to change this a bit, we basically need these functions:

```rust
fn try_to_decoded(&self) -> impl Iterator<alloy_rlp::Result<OpTxEnvelope>> + '_ 

fn try_to_recovered(&self) -> impl Iterator<Result<Recovered<OpTxEnvelope>>, SignatureError> + '_ 
```

SignatureError as a way to create an error via `Box<dyn Err>` in `std` and if not we can simply use String



```

```

crates/rpc-types-engine/src/attributes.rs
```diff
@@ -56,55 +55,80 @@ impl OpPayloadAttributes {
...
                 .map(|recovered| WithEncoded::new(tx_bytes.clone(), recovered))
-                .ok_or_else(|| {
-                    Box::<dyn error::Error>::from(format!(
-                        "Recovery error: failed to decode transaction bytes: {:?}",
-                        tx_bytes
-                    ))
-                })
+                .ok_or(SignatureError::FromBytes(
+                    "Recovery error: failed to decode transaction bytes",
+                ))
```
mattsse, line 125: this one can reuse `decoded_transactions_with_encoded` and map the items accordinly

crates/rpc-types-engine/src/attributes.rs
```diff
@@ -56,55 +55,80 @@ impl OpPayloadAttributes {
...
+
+            Ok(Recovered::new_unchecked(recovered_op_tx, recovered.signer()))
+        })
     }
 
-    /// Returns an iterator over `Result<WithEncoded<OpTxEnvelope>>`
+    /// Helper: returns first successfully recovered transaction, if any.
+    pub fn try_into_recovered(&self) -> Option<Recovered<OpTxEnvelope>> {
+        self.try_to_recovered().find_map(Result::ok)
+    }
```
mattsse, line 97: this one we can remove

crates/rpc-types-engine/src/attributes.rs
```diff
@@ -56,55 +55,80 @@ impl OpPayloadAttributes {
...
+                .try_into_recovered()
+                .map_err(|_| SignatureError::FromBytes("Failed to recover signature"))?;
+
+            let recovered_op_tx = OpTxEnvelope::try_from_eth_envelope(recovered.inner().clone())
+                .map_err(|_| {
+                    SignatureError::FromBytes("Failed to convert back from eth envelope")
+                })?;
+
+            Ok(Recovered::new_unchecked(recovered_op_tx, recovered.signer()))
+        })
```
mattsse, line 91: this one can reuse `recovered_transactions_with_encoded` and map

crates/rpc-types-engine/src/attributes.rs
```diff
@@ -56,55 +55,80 @@ impl OpPayloadAttributes {
...
+    /// Decode each transaction and return an iterator of decoded `OpTxEnvelope`s.
+    /// Returns iterator over successfully decoded OpTxEnvelope.
+    pub fn try_to_decoded(
+        &self,
+    ) -> impl Iterator<Item = Result<OpTxEnvelope, SignatureError>> + '_ {
+        self.transactions.iter().flatten().map(|tx_bytes| {
+            OpTxEnvelope::decode_2718(&mut tx_bytes.as_ref())
+                .map_err(|_| SignatureError::FromBytes("Failed to decode OpTxEnvelope"))
+        })
+    }
```
mattsse, line 67: this can return `alloy_rlp::Result`

mattsse (changes requested):
getting closer now we start reusing and mapping iterator fns

crates/rpc-types-engine/src/attributes.rs
Rimeeeeee: `recovered_transactions_with_encoded` is using this fn..do you want me to move the logic from here to the other fn and then refactor??

Rimeeeeee:
@mattsse i have refactored the logic of `decoded_transactions_with_encoded` by enabling `try_to_decoded` in it and rest.. lmk wht u think

Rimeeeeee:
@mattsse is this ok?

mattsse (approved):
nice work, I only fixed the k256 feature that was causing ci failures and renamed the transaction.

the zip combination is neat

PR #497, opened 2025-04-14
fix(rpc-types-engine): Rename op payload sidecar field

emhane:
Fixes naming, replacing `canyon` with `ecotone` for referring to cancun fields

### DIFF
--- a/crates/rpc-types-engine/src/sidecar.rs
+++ b/crates/rpc-types-engine/src/sidecar.rs
@@ -8,15 +8,15 @@ use alloy_rpc_types_engine::{
 /// Container type for all available additional `newPayload` request parameters that are not present
 /// in the [`ExecutionPayload`](alloy_rpc_types_engine::ExecutionPayload) object itself.
 ///
-/// Default is equivalent to pre-canyon, payloads v1 and v2.
+/// Default is equivalent to pre-ecotone, payloads v1 and v2.
 #[derive(Debug, Clone, Default)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct OpExecutionPayloadSidecar {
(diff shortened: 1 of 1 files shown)


crates/rpc-types-engine/src/sidecar.rs
```diff
@@ -28,45 +28,45 @@ impl OpExecutionPayloadSidecar {
...
-    pub fn v4(canyon: CancunPayloadFields, isthmus: PraguePayloadFields) -> Self {
-        Self { canyon: canyon.into(), isthmus: isthmus.into() }
+    pub fn v4(ecotone: CancunPayloadFields, isthmus: PraguePayloadFields) -> Self {
+        Self { ecotone: ecotone.into(), isthmus: isthmus.into() }
     }
 
     /// Returns a reference to the [`CancunPayloadFields`].
-    pub const fn canyon(&self) -> Option<&CancunPayloadFields> {
-        self.canyon.as_ref()
+    pub const fn ecotone(&self) -> Option<&CancunPayloadFields> {
```
mattsse, line 63: these are breaking

mattsse (changes requested):
could we do this without breaking the api here

definitely worth documenting this better but I'd prefer if we could keep the pub canyon function so we dont break it in reth and can release this as patch.

fine with duplicating the ecotone fns and maybe marking the canyon fns as deprecated

crates/rpc-types-engine/src/sidecar.rs
emhane: we care about breaking public api now 😮 i see, i see

sure, can do

mattsse: if we break stuff, it requires more work and slows down rollouts

mattsse (approved):

PR #514, opened 2025-05-08
feat: made TxDeposit's mint field non-optional

### LINKED ISSUE #513: [Feature] Make `TxDeposit`'s `mint` field non-optional
### Component

consensus

Currently, `TxDeposit`'s `mint` field is optional. This creates ambiguity in several places, namely with serialization.

In practice, `None` is treated as a functional equivalent of `Some(0)` across the board, so we can change this type to just be a `u128`.

### Additional context

_No response_

Rimeeeeee:
fixes #513

### DIFF
--- a/crates/consensus/src/transaction/deposit.rs
+++ b/crates/consensus/src/transaction/deposit.rs
@@ -28,8 +28,8 @@ pub struct TxDeposit {
     #[cfg_attr(feature = "serde", serde(default, skip_serializing_if = "TxKind::is_create"))]
     pub to: TxKind,
     /// The ETH value to mint on L2.
-    #[cfg_attr(feature = "serde", serde(default, with = "alloy_serde::quantity::opt"))]
-    pub mint: Option<u128>,
+    #[cfg_attr(feature = "serde", serde(default, with = "alloy_serde::quantity"))]
+    pub mint: u128,
     ///  The ETH value to send to the recipient account.
     pub value: U256,
     /// The gas limit for the L2 transaction.
@@ -72,9 +72,9 @@ impl TxDeposit {
             to: Decodable::decode(buf)?,
             mint: if *buf.first().ok_or(DecodeError::InputTooShort)? == EMPTY_STRING_CODE {
                 buf.advance(1);
-                None
+                0
             } else {
-                Some(Decodable::decode(buf)?)
+                Decodable::decode(buf)?
             },
             value: Decodable::decode(buf)?,
             gas_limit: Decodable::decode(buf)?,
@@ -110,7 +110,7 @@ impl TxDeposit {
         self.source_hash.length()
             + self.from.length()
             + self.to.length()
-            + self.mint.map_or(1, |mint| mint.length())
+            + self.mint.length()
             + self.value.length()
             + self.gas_limit.length()
             + self.is_system_transaction.length()
@@ -123,10 +123,10 @@ impl TxDeposit {
         self.source_hash.encode(out);
         self.from.encode(out);
         self.to.encode(out);
-        if let Some(mint) = self.mint {
-            mint.encode(out);
-        } else {
+        if self.mint == 0 {
             out.put_u8(EMPTY_STRING_CODE);
+        } else {
+            self.mint.encode(out);
         }
         self.value.encode(out);
         self.gas_limit.encode(out);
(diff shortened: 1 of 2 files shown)


crates/consensus/src/transaction/deposit.rs
```diff
@@ -123,10 +123,10 @@ impl TxDeposit {
...
         self.from.encode(out);
         self.to.encode(out);
-        if let Some(mint) = self.mint {
-            mint.encode(out);
-        } else {
+        if self.mint == 0 {
             out.put_u8(EMPTY_STRING_CODE);
+        } else {
+            self.mint.encode(out);
         }
```
mattsse, line 130: I think we no longer have to do this because rlp(u128::Zero) is the same as EMPTY_STRING_CODE

crates/consensus/src/transaction/deposit.rs
```diff
@@ -370,7 +370,7 @@ impl DepositTransaction for TxDeposit {
 
     #[inline]
     fn mint(&self) -> Option<u128> {
-        self.mint
+        Some(self.mint)
     }
```
mattsse, line 374: we can also remove the Option for this

crates/consensus/src/transaction/envelope.rs
```diff
@@ -972,7 +972,7 @@ mod tests {
 
         let tx = OpTxEnvelope::decode_2718(&mut b[..].as_ref()).unwrap();
         let deposit = tx.as_deposit().unwrap();
-        assert!(deposit.mint.is_none());
+        assert!(deposit.mint == 0);
```
mattsse, line 975:
```suggestion
        assert_eq!(deposit.mint, 0);
```

crates/consensus/src/transaction/deposit.rs
```diff
@@ -72,9 +72,9 @@ impl TxDeposit {
             to: Decodable::decode(buf)?,
             mint: if *buf.first().ok_or(DecodeError::InputTooShort)? == EMPTY_STRING_CODE {
                 buf.advance(1);
-                None
+                0
             } else {
-                Some(Decodable::decode(buf)?)
+                Decodable::decode(buf)?
             },
```
mattsse, line 78: I believe we can now also simplify this to just `Decodable::decode(buf)?`

because `EMPTY_STRING_CODE` is the same as rlp(0)

mattsse (changes requested):
great

maybe let's add a few rlp roundtrip tests as well

you can get raw deposit txs from basescan for example

the first tx in a block is awlays a deposit, for example:

https://basescan.org/getRawTx?tx=0xa68157af562dec44406bf9334381cf7e46aab0dd3f6495d99644e02b6fed6f09

crates/consensus/src/transaction/deposit.rs
```diff
@@ -573,6 +562,30 @@ mod tests {
...
+    #[test]
+    fn test_deposit_tx_roundtrip() {
+        let raw_txs = [
+            "7ef8f8a0871ec5fb6afe7e5ae950bbb4cfd7d7cb277b413e67da806d50834a814b14c9f494deaddeaddeaddeaddeaddeaddeaddeaddead00019442000000000000000000000000000000000000158080830f424080b8a4440a5e20000008dd00101c12000000000000000400000000681c941f0000000001566261000000000000000000000000000000000000000000000000000000005f629c020000000000000000000000000000000000000000000000000000000000000001937badfbcce566e0ba932a3f7659644aa0c6ef019541d3134a1d8cb9f84d45c70000000000000000000000005050f69a9786f081509234f1a7f4684b5e5b76c9",
+        ];
+
+        for raw_tx_hex in raw_txs {
+            let raw_tx = hex::decode(raw_tx_hex).unwrap();
+
+            let tx = TxDeposit::typed_decode(raw_tx[0], &mut &raw_tx[1..]).unwrap();
```
mattsse, line 574: let's use the decode_2718 impl here

mattsse (changes requested):
last nit, then this is g2m

will block merge shortly

cc @clabby for vis

clabby (approved):
Looks good, thanks for picking this up.

Will defer to @mattsse to merge when he's ready.

mattsse (approved):

Issue #517, opened 2025-05-12
Migrate type from reth


emhane:
Migrate this file from reth into `op_alloy_consensus::interop` so it can be used by kona and op-reth
https://github.com/paradigmxyz/reth/blob/82f458081a05b911b64a3a82d8a7d8270a1abc6a/crates/optimism/txpool/src/supervisor/errors.rs#L1-L129

PR #518, opened 2025-05-12
chore(super-consensus): Migrate `InvalidInboxEntry` from reth

emhane:
Closes https://github.com/alloy-rs/op-alloy/issues/517

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,23 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.15.7](https://github.com/alloy-rs/op-alloy
+/releases/tag/v0.15.7) - 2025-05-13
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


crates/consensus/Cargo.toml
```diff
@@ -20,14 +20,15 @@ alloy-rlp.workspace = true
 alloy-eips.workspace = true
 alloy-consensus.workspace = true
 alloy-primitives = { workspace = true, features = ["rlp"] }
+alloy-json-rpc.workspace = true
```
mattsse, line 23: I think this one could be problematic because:

https://github.com/alloy-rs/alloy/blob/e94b8d4439fd5e28da354d185e68623ec0b1db2e/crates/json-rpc/Cargo.toml#L28-L29

mattsse (changes requested):
should we maybe move this to the rpc-types crate instead?

crates/consensus/Cargo.toml
emhane: yeah we can do that :)

emhane:
`once_cell` is failing CI, we have had issue with this dep before right? which version should I pin it to?

emhane:
@mattsse ping for re-review, shrank the scope !

mattsse (approved):
nice

PR #525, opened 2025-05-20
Yash/rm op tx req

yash-atreya:
## Motivation

Ref https://github.com/alloy-rs/alloy/issues/2478

`OpTransactionRequest` is a wrapper around `TransactionRequest`.

It doesn't implement `TransactionBuilder4844` (It shouldn't). But this makes it incompatible with `ProviderBuilder` default constructor `new`

## Solution

- Remove OpTransactionRequest
- Move related `From` conversions to alloy_compat
- Use TransactionRequest in network impl

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


.github/workflows/ci.yml
```diff
@@ -85,7 +85,7 @@ jobs:
           cache-on-failure: true
       - name: cargo hack
         run: |
-          cargo hack build --workspace --target wasm32-unknown-unknown --exclude op-alloy-network --exclude op-alloy-rpc-jsonrpsee --exclude op-alloy-provider
+          cargo hack build --workspace --target wasm32-unknown-unknown --exclude op-alloy-network --exclude op-alloy-rpc-jsonrpsee --exclude op-alloy-provider --exclude op-alloy-consensus
```
mattsse, line 88: this change shouldn't be necessary

crates/consensus/Cargo.toml
```diff
@@ -47,7 +47,7 @@ arbitrary = { workspace = true, features = ["derive"] }
 alloy-primitives = { workspace = true, features = ["rand", "arbitrary"] }
 
 [features]
-default = ["std"]
+default = ["std", "alloy-compat"]
```
mattsse, line 50: why do we need this by default now

crates/consensus/src/alloy_compat.rs
```diff
@@ -92,7 +142,8 @@ mod tests {
         let unknown_tx_envelope: UnknownTxEnvelope = serde_json::from_str(deposit).unwrap();
 
         let _deposit: TxDeposit = unknown_tx_envelope.try_into().unwrap();
-        dbg!(&_deposit);
+        #[cfg(feature = "std")]
+        dbg!(_deposit);
```
mattsse, line 146: this can be removed entirely

mattsse (changes requested):
smol nits

I think this is reasonable, no reason to introduce a new txrequest type here really because most of the server apis are restricted to TransactionRequest, so even if this does allow configuring 4844 fields, this seems fine

crates/network/src/lib.rs
```diff
@@ -43,120 +44,153 @@ impl Network for Optimism {
...
+
+        // chain_id and from may be none.
+        let common = self.gas.is_some() && self.nonce.is_some();
+
+        let legacy = self.gas_price.is_some();
+        let eip2930 = legacy && self.access_list.is_some();
+
+        let eip1559 = self.max_fee_per_gas.is_some() && self.max_priority_fee_per_gas.is_some();
+
+        let eip4844 = eip1559 && self.sidecar.is_some() && self.to.is_some();
```
emhane, line 175: this can just return `false` early and can be checked as first line of this method to bail as early as possible. imo should also log error or panic in debug mode.

crates/network/src/lib.rs
```diff
@@ -175,16 +221,52 @@ impl TransactionBuilder<Optimism> for OpTransactionRequest {
...
+                    nonce: tx.nonce,
+                    gas_limit: tx.gas_limit,
+                    max_priority_fee_per_gas: tx.max_priority_fee_per_gas,
+                    max_fee_per_gas: tx.max_fee_per_gas,
+                    to: TxKind::Call(tx.to),
+                    value: tx.value,
+                    access_list: tx.access_list,
+                    input: tx.input,
+                }))
+            }
```
emhane, line 264: this doesn't seem right, this should return an error instead and possibly even panic in debug mode. eip4844 transactions shouldn't be accepted in any part of the op tx processing pipeline.

crates/network/src/lib.rs
```diff
@@ -43,120 +44,153 @@ impl Network for Optimism {
...
         match ty {
             OpTxType::Deposit => Err(vec!["not implemented for deposit tx"]),
             _ => {
                 let ty = TxType::try_from(ty as u8).unwrap();
-                self.as_ref().complete_type(ty)
+                match ty {
+                    TxType::Legacy => self.complete_legacy(),
+                    TxType::Eip2930 => self.complete_2930(),
+                    TxType::Eip1559 => self.complete_1559(),
+                    TxType::Eip4844 => Err(vec!["cannot build eip4844 tx"]),
```
emhane, line 149: this is the behaviour we want for 4844 txns

crates/network/src/lib.rs
```diff
@@ -43,120 +44,153 @@ impl Network for Optimism {
...
             }
         }
     }
 
     fn can_submit(&self) -> bool {
-        self.as_ref().can_submit()
+        // value and data may be None. If they are, they will be set to default.
+        // gas fields and nonce may be None, if they are, they will be populated
+        // with default values by the RPC server
+        self.from.is_some()
```
emhane, line 160: this comment doesn't make a lot of sense to me here

emhane (changes requested):

crates/network/src/lib.rs
yash-atreya: It's equivalent to `build_typed_tx` currently on main https://github.com/alloy-rs/op-alloy/blob/main/crates/rpc-types/src/transaction/request.rs#L104-L117 which doesn't return an error.

crates/network/src/lib.rs
yash-atreya: [98e4dbb](https://github.com/alloy-rs/op-alloy/pull/525/commits/98e4dbb64b7e55f6a2d027727779c96a69f95726)

crates/network/src/lib.rs
yash-atreya: [98e4dbb](https://github.com/alloy-rs/op-alloy/pull/525/commits/98e4dbb64b7e55f6a2d027727779c96a69f95726)

crates/network/src/lib.rs
emhane: ok can be addressed in follow up

crates/consensus/src/transaction/envelope.rs
```diff
@@ -132,6 +133,18 @@ impl TryFrom<OpTxEnvelope> for TxEnvelope {
...
 }
 
+impl From<OpTxEnvelope> for TransactionRequest {
+    fn from(value: OpTxEnvelope) -> Self {
+        match value {
+            OpTxEnvelope::Eip2930(tx) => tx.into_parts().0.into(),
+            OpTxEnvelope::Eip1559(tx) => tx.into_parts().0.into(),
+            OpTxEnvelope::Eip7702(tx) => tx.into_parts().0.into(),
+            OpTxEnvelope::Deposit(tx) => tx.into_inner().into(),
+            _ => Default::default(),
```
emhane, line 143: let's not use a wildcard here, makes it harder to maintain

emhane (approved):

crates/network/src/lib.rs
```diff
@@ -175,16 +220,52 @@ impl TransactionBuilder<Optimism> for OpTransactionRequest {
...
+            }
+            EthereumTypedTransaction::Eip2930(tx) => {
+                let tx = OpTypedTransaction::Eip2930(tx);
+                Ok(tx)
+            }
+            EthereumTypedTransaction::Eip1559(tx) => {
+                let tx = OpTypedTransaction::Eip1559(tx);
+                Ok(tx)
+            }
+            EthereumTypedTransaction::Eip4844(tx) => {
```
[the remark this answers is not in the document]
yash-atreya: Can remove it. It is being handled on main. Just ported it here. See: https://github.com/alloy-rs/op-alloy/pull/525#discussion_r2100427647
> It's equivalent to build_typed_tx currently on main https://github.com/alloy-rs/op-alloy/blob/main/crates/rpc-types/src/transaction/request.rs#L104-L117 which doesn't return an error.

crates/consensus/src/transaction/deposit.rs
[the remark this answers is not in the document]
yash-atreya: It was implemeted for `OpTransactionRequest`: https://github.com/alloy-rs/op-alloy/blob/b454f665155732dcc2cfbdfadde3c0a473bd69a0/crates/rpc-types/src/transaction/request.rs#L123

After removing that, implemented similar conversion for `TransactionRequest`

crates/network/src/lib.rs
```diff
@@ -43,120 +44,152 @@ impl Network for Optimism {
...
     fn set_access_list(&mut self, access_list: AccessList) {
-        self.as_mut().set_access_list(access_list);
+        self.access_list = Some(access_list);
     }
 
     fn complete_type(&self, ty: OpTxType) -> Result<(), Vec<&'static str>> {
         match ty {
             OpTxType::Deposit => Err(vec!["not implemented for deposit tx"]),
             _ => {
                 let ty = TxType::try_from(ty as u8).unwrap();
```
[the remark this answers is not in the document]
yash-atreya: Fixed [4287ab2](https://github.com/alloy-rs/op-alloy/pull/525/commits/4287ab27a1982338bff31daaa0ded1aed4cb4979)

crates/network/src/lib.rs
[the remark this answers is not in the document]
yash-atreya: Fixed [5b35d0e](https://github.com/alloy-rs/op-alloy/pull/525/commits/5b35d0e7f20698a50277bf0f6c49a5f9e2b9e7e0)

crates/consensus/Cargo.toml
```diff
@@ -23,7 +23,7 @@ alloy-primitives = { workspace = true, features = ["rlp"] }
 
 # compat
 alloy-network = { workspace = true, optional = true }
-alloy-rpc-types-eth = { workspace = true, optional = true }
+alloy-rpc-types-eth.workspace = true
```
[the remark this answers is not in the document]
yash-atreya: This is due to the impl From for `TransactionRequest` conversion being moved to their respective files after removal of `OpTransactionRequest`. 

These conversions are: 

1. `impl From<TxDeposit> for TransactionRequest` in deposit.rs

2. `impl From<OpTxEnvelope> for TransactionRequest` in envelope.rs

3. `impl From<OpTypedTransaction> for TransactionRequest` in typed.rs 

I can gate them behind `alloy-compat`?

PR #527, opened 2025-05-25
feat: add meta helpers

mattsse:
these will allow us to offload this impl entirely

https://github.com/paradigmxyz/reth/blob/4df1425fcf860572331c682cf868d0a94129536f/crates/optimism/rpc/src/eth/transaction.rs#L93-L97

ref https://github.com/paradigmxyz/reth/issues/16451

### DIFF
--- a/crates/consensus/src/transaction/mod.rs
+++ b/crates/consensus/src/transaction/mod.rs
@@ -1,4 +1,4 @@
-//! Tramsaction types for Optimism.
+//! Transaction types for Optimism.
 
 mod deposit;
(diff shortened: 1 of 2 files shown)


crates/consensus/src/transaction/mod.rs
```diff
@@ -1,4 +1,4 @@
-//! Tramsaction types for Optimism.
```
clabby, line 1: 😄 
GettyImages-1576504838-e1729846870500 copy

clabby (approved):
Makes sense

PR #529, opened 2025-05-26
feat: added Transaction conversion from consensus for rpc 

### LINKED ISSUE #528: [Feature] Add Transaction conversion from consensus for rpc
### Component

op-alloy

with https://github.com/alloy-rs/op-alloy/pull/527 we now have a way to offload this entire reth code:

https://github.com/paradigmxyz/reth/blob/4df1425fcf860572331c682cf868d0a94129536f/crates/optimism/rpc/src/eth/transaction.rs#L93-L97

if we make this a constructor fn of the rpc tx type with OpTransactionInfo as input + `Recovered<OpTxEnvelope>` as input

https://github.com/alloy-rs/op-alloy/blob/f5858c6cbe11567b7c550e7f3755cb04dee8af56/crates/rpc-types/src/transaction.rs#L19-L19

## TODO
* port reth code to rpc optransaction

### Additional context

_No response_

Soubhik-10:
closes #528

### DIFF
--- a/crates/rpc-types/src/transaction.rs
+++ b/crates/rpc-types/src/transaction.rs
@@ -1,6 +1,6 @@
 //! Optimism specific types related to transactions.
 
-use alloy_consensus::{Transaction as _, Typed2718};
+use alloy_consensus::{Transaction as _, Typed2718, transaction::Recovered};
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm!

we can do the same for the alloy one

PR #535, opened 2025-06-05
feat(interop): Rename `InvalidInboxEntry`->`SuperchainDAError`

emhane:
Renames `InvalidInboxEntry` to `SuperchainDAError`.

Some of these error variants occur also on updating superchain state. The name `InvalidInboxEntry` is to specialised for `supervisor_checkAccessList`.

Adds feature `jsonrpsee` to `op-alloy-rpc-types` and feature gated conversion to `jsonrpsee::ErrorObjectOwned`.

### DIFF
--- a/crates/rpc-types/src/error.rs
+++ b/crates/rpc-types/src/error.rs
@@ -105,6 +105,8 @@ pub enum SuperchainDAError {
 #[cfg(feature = "jsonrpsee")]
 impl From<SuperchainDAError> for jsonrpsee::types::ErrorObjectOwned {
     fn from(err: SuperchainDAError) -> Self {
+        use crate::alloc::string::ToString;
+        
         jsonrpsee::types::ErrorObjectOwned::owned(err as i32, err.to_string(), None::<()>)
     }
 }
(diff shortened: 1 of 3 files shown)


crates/rpc-types/src/error.rs
```diff
@@ -101,3 +101,19 @@ pub enum InvalidInboxEntry {
...
+}
+
+#[cfg(feature = "jsonrpsee")]
+impl TryFrom<jsonrpsee::types::ErrorObjectOwned> for SuperchainDAError {
+    type Error = derive_more::TryFromReprError<i32>;
+
+    fn try_from(err: jsonrpsee::types::ErrorObjectOwned) -> Result<Self, Self::Error> {
+        err.code().try_into()
+    }
+}
```
mattsse, line 119: do we need this conversion?

mattsse (commented):
makes sense, one question

crates/rpc-types/src/error.rs
emhane: good catch, no not really since reth is using alloy generic rpc error I see now

PR #539, opened 2025-06-10
feat(rpc-types-engine): OpExecutionPayload Wrapper

refcell:
### Description

A thin wrapper around `OpExecutionPayload` that includes the parent beacon block root that provides a rust-equivalent to [`ExecutionPayloadEnvelope`](https://github.com/ethereum-optimism/optimism/blob/7c812674f9e2068949a152f44fde13fcb93b2c45/op-service/eth/types.go#L233).

This is required for a serialized rpc type in the op admin api to post unsafe payloads.

### DIFF
--- a/crates/rpc-types-engine/src/envelope.rs
+++ b/crates/rpc-types-engine/src/envelope.rs
@@ -15,6 +15,7 @@ use alloy_rpc_types_engine::{
 /// A thin wrapper around [`OpExecutionPayload`] that includes the parent beacon block root.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 pub struct OpExecutionPayloadEnvelope {
     /// The execution payload.
     pub payload: OpExecutionPayload,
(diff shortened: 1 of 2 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #539 👈
* `main`

crates/rpc-types-engine/src/envelope.rs
```diff
@@ -12,6 +12,25 @@ use alloy_rpc_types_engine::{
...
 };
 
+/// A thin wrapper around [`OpExecutionPayload`] that includes the parent beacon block root.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct OpExecutionPayloadEnvelope {
+    /// The execution payload.
+    pub payload: OpExecutionPayload,
+    /// The parent beacon block root, if any.
+    pub parent_beacon_block_root: Option<B256>,
```
mattsse, line 22: this needs camelcase rename, right?

mattsse (changes requested):

emhane (dismissed):
this type exists as `OpExecutionData`

refcell:
> this type exists as `OpExecutionData`

That type is not the same. `OpExecutionData` does not `#[serde(flatten)]` the `sidecar` field so its serialized json representation is different than the type introduced in this pr.

mattsse (approved):
makes sense, given that op-geth also has this type

PR #540, opened 2025-06-10
fix: redirect to op-alloy-provider crate

AskMeLuoryyy:
## Motivation

on this page is bug - https://alloy-rs.github.io/op-alloy/starting.html 
image

## Solution

due to this line https://github.com/alloy-rs/op-alloy/blob/main/book/src/links.md?plain=1#L10 we need to do this change. 
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


emhane (approved):
thanks

PR #542, opened 2025-06-11
feat(rpc): Replace wrapped `OpTxEnvelope` with a generic in the `Transaction` RPC response

RomanHodulak:
## Motivation
Extending the transaction RPC response requires to rewrite this entire type.

## Solution
Replace the `OpTxEnvelope` with a generic parameter constrained to `alloy_consensus::Transaction` and `SignerRecoverable`.

Allows reusing `Transaction` RPC response object with custom transactions.

Simplifies the creation of custom nodes.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -659,27 +659,6 @@ impl alloy_consensus::transaction::SignerRecoverable for OpTxEnvelope {
     }
 }
 
-#[cfg(not(feature = "k256"))]
-impl alloy_consensus::transaction::SignerRecoverable for OpTxEnvelope {
-    fn recover_signer(&self) -> Result<Address, alloy_consensus::crypto::RecoveryError> {
-        match self {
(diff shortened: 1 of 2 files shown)


mattsse (commented):
I see that we do something funny with the from field during serde hence the requirement for signerrecoverable

since this is for deposit only, I think we could instead introduce a dedicated trait like OpEnvelope that returns `Option<&TxDeposit>`?

wdyt @klkvr

klkvr:
> since this is for deposit only, I think we could instead introduce a dedicated trait like OpEnvelope that returns `Option<&TxDeposit>`?

yeah I think we could add it to https://github.com/paradigmxyz/reth/blob/0284420012dddae3c699f9826b66b0ac37dfdf9e/crates/optimism/primitives/src/transaction/mod.rs#L15-L20 and move it here

mattsse:
sg!
opening an issue for this

PR #544, opened 2025-06-12
feat(rpc-types-engine): SSZ Encoding

refcell:
### Description

In order to construct a signature to add to an `OpNetworkPayloadEnvelope`, the op payload has to be serialized as ssz.
The serialized ssz bytes are then signed over using a local signer.
Along with constructing the payload hash, the signature is used to construct the `OpNetworkPayloadEnvelope`.

Once the `OpNetworkPayloadEnvelope` is constructed with the signature over the ssz encoded payload bytes,
the `OpNetworkPayloadEnvelope` is encoded and sent over the wire through op p2p gossip.


clabby (approved):
LGTM

PR #549, opened 2025-06-16
feat(consensus): Add `as_deposit` into `OpTransaction` trait

RomanHodulak:
Follow-up on #548 

## Motivation

`OpTransaction` trait can tell if its deposit, but that is not very useful if I also need to extract it.

## Solution

Add `as_deposit` function into the trait for the extraction of `TxDeposit`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -51,13 +51,20 @@ pub enum OpTxEnvelope {
 ///
 /// Compared to Ethereum it can tell whether the transaction is a deposit.
 pub trait OpTransaction {
-    /// Returns true if the transaction is a deposit.
+    /// Returns `true` if the transaction is a deposit.
     fn is_deposit(&self) -> bool;
+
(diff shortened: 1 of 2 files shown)


crates/consensus/src/transaction/envelope.rs
```diff
@@ -51,27 +51,48 @@ pub enum OpTxEnvelope {
 ///
 /// Compared to Ethereum it can tell whether the transaction is a deposit.
 pub trait OpTransaction {
-    /// Returns true if the transaction is a deposit.
+    /// An associated deposit transaction variant.
+    type Deposit: DepositTransaction;
```
mattsse, line 55: hmm do we need this AT, I think it's safe to assume that for op the txdeposit type will always be the same, because this one is specced by opstack?

this wouldn't solve the from field access problem for #542

the trait DepositTransaction is probably not very useful

PR #553, opened 2025-06-17
feat(rpc): Convert into RPC transaction for generic inner `OpTransaction`

RomanHodulak:
## Motivation

Analogous to alloy-rs/alloy#2591

The RPC transaction can be created from a generic `alloy_consensus::Transaction`. However, this does not cover the case when the RPC inner transaction needs to be converted.

## Solution

Add another generic transaction as the input for the conversion, with a `From` bound and a `convert` call.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/transaction.rs
+++ b/crates/rpc-types/src/transaction.rs
@@ -35,7 +35,11 @@ pub struct Transaction<T = OpTxEnvelope> {
 
 impl<T: OpTransaction + TransactionTrait> Transaction<T> {
     /// Converts a consensus `tx` with an additional context `tx_info` into an RPC [`Transaction`].
-    pub fn from_transaction(tx: Recovered<T>, tx_info: OpTransactionInfo) -> Self {
+    pub fn from_transaction<TxIn>(tx: Recovered<TxIn>, tx_info: OpTransactionInfo) -> Self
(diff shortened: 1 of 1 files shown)


crates/rpc-types/src/transaction.rs
```diff
@@ -35,7 +35,11 @@ pub struct Transaction<T = OpTxEnvelope> {
 
 impl<T: OpTransaction + TransactionTrait> Transaction<T> {
     /// Converts a consensus `tx` with an additional context `tx_info` into an RPC [`Transaction`].
-    pub fn from_transaction(tx: Recovered<T>, tx_info: OpTransactionInfo) -> Self {
+    pub fn from_transaction<TxIn>(tx: Recovered<TxIn>, tx_info: OpTransactionInfo) -> Self
```
mattsse, line 38: isn't this the same as calling this with `tx.convert()`?

RomanHodulak: It is. But more importantly, we're facing the same issue as mentioned in a comment of alloy-rs/alloy#2591

RomanHodulak:
Closing, see the discussion.

PR #569, opened 2025-07-15
refactor(consensus): simplify is_deposit method in OpTxType

reallesee:
Replace matches! macro with direct equality comparison for better readability and performance. This change maintains the same functionality while making the code more idiomatic and potentially allowing for better compiler optimizations

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


emhane:
not sure this is actually simpler, have you

PR #572, opened 2025-07-17
fix(rpc-types-engine): Fix `OpExecutionPayloadEnvelope::payload_hash`

clabby:
## Overview

Fixes the `OpExecutionPayloadEnvelope::payload_hash` function to be in-line with the [spec](https://specs.optimism.io/protocol/rollup-node-p2p.html#block-signatures).


emhane (approved):
would be great if we can add that spec link to the docs of this method

mattsse (approved):

PR #578, opened 2025-07-21
fix(rpc-types-engine): `OpExecutionPayloadEnvelope` serialization fix

clabby:
## Overview

Fixes the serialization of the `OpExecutionPayloadEnvelope` type, changing the `payload` field -> `execution_payload`. ref: https://github.com/ethereum-optimism/optimism/blob/HEAD/op-service/eth/types.go#L235-L235


mattsse (approved):
sigh

PR #580, opened 2025-07-28
[jovian] extend `extraData` to include minBaseFee in EIP1559Params

wlawt:
From: https://github.com/wlawt/op-alloy/pull/1

## Motivation

This PR adds the necessary helper functions to encode/decode the [configurable minimum base fee](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/minimum-base-fee.md) feature to op-alloy. 

Specifically, this extends the `extraData` section in the block header to include an extra uint64 for this addition. The default value would be zero, as if there was no minimum base fee to begin with if the configurable minimum base fee wasn't enabled. 

By introducing a minimum base fee, it reduces the chances of experiencing priority fee auctions which occur when the block is almost at the gas limit, as well as cases where batcher sequencer throttling is happening where it intentionally processes fewer transactions to avoid a backlog of batches.

## Solution

By setting a minimum base fee, we reduce the time and the number of doublings needed. See the table on the time saved on [OP Mainnet in minutes](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/minimum-base-fee.md#minimum-base-fee-in-block-header).

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/eip1559.rs
+++ b/crates/consensus/src/eip1559.rs
@@ -131,6 +131,12 @@ pub enum EIP1559ParamError {
     /// Extra data is not the correct length.
     #[error("Extra data is not the correct length")]
     InvalidExtraDataLength,
+    /// Minimum base fee must be None before Jovian.
+    #[error("Minimum base fee must be None before Jovian")]
+    MinBaseFeeMustBeNone,
+    /// Minimum base fee cannot be None after Jovian.
+    #[error("Minimum base fee cannot be None after Jovian")]
+    MinBaseFeeNotSet,
 }
 
 #[cfg(test)]
@@ -153,7 +159,7 @@ mod tests {
     }
 
     #[test]
-    fn test_get_extra_data_min_base_fee() {
+    fn test_get_extra_data_post_jovian() {
         let eip_1559_params = B64::from_str("0x0000000800000008").unwrap();
         let extra_data = encode_jovian_extra_data(eip_1559_params, BaseFeeParams::new(80, 60), 257);
         assert_eq!(
@@ -163,7 +169,7 @@ mod tests {
     }
 
     #[test]
-    fn test_get_extra_data_min_base_fee_default() {
+    fn test_get_extra_data_post_jovian_default() {
         let eip_1559_params = B64::ZERO;
         let extra_data = encode_jovian_extra_data(eip_1559_params, BaseFeeParams::new(80, 60), 0);
         // check the version byte is 1 and the min_base_fee is 0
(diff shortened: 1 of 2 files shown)


wlawt:
cc: @niran

crates/rpc-types/src/genesis.rs
```diff
@@ -262,7 +263,7 @@ mod tests {
                     holocene_time: Some(0),
                     isthmus_time: None,
```
emhane, line 264:
```suggestion
                    isthmus_time: Some(0),
```
isthmus should be enabled if jovian is enabled

crates/rpc-types/src/genesis.rs
```diff
@@ -242,7 +242,8 @@ mod tests {
           "ecotoneTime": 0,
           "fjordTime": 0,
           "graniteTime": 0,
-          "holoceneTime": 0
+          "holoceneTime": 0,
```
emhane, line 245:
```suggestion
          "holoceneTime": 0,
          "isthmusTime": 0,
```
isthmus should be enabled if jovian is enabled

wlawt:
cc: @theochap

theochap:
@wlawt I am going to start reviewing that one

crates/consensus/src/eip1559.rs
```diff
@@ -3,6 +3,34 @@
...
 use alloy_primitives::{B64, Bytes};
 
+const HOLOCENE_EXTRA_DATA_VERSION_BYTE: u8 = 0;
+const JOVIAN_EXTRA_DATA_VERSION_BYTE: u8 = 1;
+
+/// Encodes the `eip1559` parameters for the payload.
+fn encode_eip_1559_params(
+    eip_1559_params: B64,
+    default_base_fee_params: BaseFeeParams,
+    extra_data: &mut [u8],
```
theochap, line 13: nit: Let's use the explicit [u8; 9] type instead to have type safety (the compiler will complain if you try to assign outside the array bounds)

crates/consensus/src/eip1559.rs
```diff
@@ -96,4 +145,25 @@ mod tests {
...
+    #[test]
+    fn test_get_extra_data_min_base_fee_default() {
+        let eip_1559_params = B64::ZERO;
+        let extra_data = encode_jovian_extra_data(eip_1559_params, BaseFeeParams::new(80, 60), 0);
+        // check the version byte is 1 and the min_base_fee is 0
+        assert_eq!(
+            extra_data.unwrap(),
+            Bytes::copy_from_slice(&[1, 0, 0, 0, 80, 0, 0, 0, 60, 0, 0, 0, 0, 0, 0, 0, 0])
+        );
+    }
```
theochap, line 168: Let's add two negative tests to ensure that the encoding/decoding fail if the length of the array is not 17 bytes

PR #606, opened 2025-11-07
feat(flashblock): commit shared flashblock data structures

0x00101010:
## Motivation

commit shared flashblock data structures

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-engine/src/flashblock/base.rs
+++ b/crates/rpc-types-engine/src/flashblock/base.rs
@@ -35,6 +35,7 @@ pub struct OpFlashblockPayloadBase {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use alloc::vec;
 
     #[test]
     #[cfg(feature = "serde")]
--- a/crates/rpc-types-engine/src/flashblock/metadata.rs
+++ b/crates/rpc-types-engine/src/flashblock/metadata.rs
@@ -22,6 +22,7 @@ pub struct OpFlashblockPayloadMetadata {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use alloc::vec::Vec;
     use alloy_consensus::{Eip658Value, Receipt};
     use alloy_primitives::{Log, address};
 
--- a/crates/consensus/src/receipts/receipt.rs
+++ b/crates/consensus/src/receipts/receipt.rs
@@ -419,6 +419,7 @@ impl From<super::OpReceiptEnvelope> for OpReceipt {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use alloc::vec;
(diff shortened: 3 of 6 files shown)


crates/rpc-types-engine/src/flashblock/metadata.rs
```diff
@@ -0,0 +1,218 @@
...
+    fn from(metadata: OpFlashblockMetadataV1) -> Self {
+        Self::V1(metadata)
+    }
+}
+
+/// Provides metadata about the block that may be useful for indexing or analysis.
+// Note: this uses mixed camel, snake case: <https://github.com/flashbots/rollup-boost/blob/dd12e8e8366004b4758bfa0cfa98efa6929b7e9f/crates/flashblocks-rpc/src/cache.rs#L31>
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct OpFlashblockMetadataV1 {
```
klkvr: this is missing this helper https://github.com/paradigmxyz/reth/blob/7997cd42832f4be41e0a20e5796f8dc194b7b137/crates/optimism/flashblocks/src/payload.rs#L78-L83

klkvr: hmm I don't think versioning and ref types make sense to me yet

what do we gain via an `OpFlashblockMetadataRef` vs `&OpFlashblockMetadata`?

klkvr: yep name changes sgtm

crates/rpc-types-engine/src/flashblock/base.rs
```diff
@@ -0,0 +1,194 @@
...
+
+use alloy_primitives::{Address, B256, Bytes, U256};
+
+/// Execution payload base envelope.
+///
+/// This enum allows for future versioning of flashblock execution payload base types.
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(untagged))]
+pub enum OpFlashblockExecutionPayloadBase {
```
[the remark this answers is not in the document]
0x00101010: I don't think they'll break per se, this is the exact same pattern as `pub enum ExecutionPayload` in `alloy-rpc-types-engine`. V1's always going to be there, all future versions are going to be optional, and we can use as_v1().v1_field, and as_v2().v2_field when exposing accesor methods.

But as discussed in below thread, personally prefer no versioning as well so no problem here changing it.

PR #609, opened 2025-11-19
chore(op-alloy): Update blob gas used serialization method

0x00101010:
## Motivation

Update blob gas used serialization method

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/receipts/receipt.rs
+++ b/crates/consensus/src/receipts/receipt.rs
@@ -520,10 +520,9 @@ pub(crate) mod serde_bincode_compat {
             data.receipt.as_receipt_mut().status = success.into();
 
             let encoded = bincode::serde::encode_to_vec(&data, bincode::config::legacy()).unwrap();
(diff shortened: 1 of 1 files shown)


crates/consensus/src/receipts/receipt.rs
```diff
@@ -519,8 +519,10 @@ pub(crate) mod serde_bincode_compat {
             // // ensure we don't have an invalid poststate variant
             data.receipt.as_receipt_mut().status = success.into();
 
-            let encoded = bincode::serialize(&data).unwrap();
-            let decoded: Data = bincode::deserialize(&encoded).unwrap();
+            let encoded = bincode::serde::encode_to_vec(&data, bincode::config::legacy()).unwrap();
```
0x00101010: old one was on bincode 1.3.3, and this repo has bincode 2.01, there's breaking changes.

This code matches other parts of the code with the same pattern

PR #610, opened 2025-11-20
feat: Use `OpReceipt` as `ReceiptEnvelope` for OP network

klkvr:
## Motivation

Based on https://github.com/alloy-rs/alloy/pull/3225

Sets `OpReceipt` as the `ReceiptResponse` for `Optimism` network implementation.

With this PR we can fully remove `OpReceiptEnvelope` but not sure if it's worth keeping it for backwards compatibility

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-engine/src/flashblock/metadata.rs
+++ b/crates/rpc-types-engine/src/flashblock/metadata.rs
@@ -16,7 +16,57 @@ pub struct OpFlashblockPayloadMetadata {
     pub new_account_balances: BTreeMap<Address, U256>,
     /// Execution receipts for all transactions in the block.
     /// Contains logs, gas usage, and other EVM-level metadata.
+    #[cfg_attr(feature = "serde", serde(deserialize_with = "deserialize_flashblock_receipts"))]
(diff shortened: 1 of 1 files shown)


crates/rpc-types-engine/src/flashblock/metadata.rs
```diff
@@ -135,8 +135,7 @@ mod tests {
             .get("0x0000000000000000000000000000000000000000000000000000000000000000")
             .unwrap();
 
-        // OpReceipt serializes as internally tagged enum
-        assert!(receipt_entry.get("Legacy").is_some());
+        assert_eq!(receipt_entry.get("type").unwrap().as_str().unwrap(), "0x0");
```
mattsse: is this changing serde of the flashblocks type?

0x00101010:
> flagging this in release notes and also
> 
> fyi @0x00101010 this allows us to clean up a bunch of receipt types but might be breaking if used as the emitting type

ack, I'll watch the release and test flashblocks side

PR #613, opened 2025-12-04
chore(rpc-types-engine): upstream `payloadID` computation to `op-alloy`

theochap:
## Description

Upstream the `payloadID` computation to `op-alloy`. This is useful to remove the dependency of `rollup-boost` onto `op-reth`, which unlocks releasing `kona` on crates.io


emhane (changes requested):
we defo want a unit test for this!

PR #618, opened 2025-12-13
refactor(net): use new `TransactionBuilder`/`NetworkTransactionBuilder` traits

lean-apple:
## Motivation

To continue https://github.com/alloy-rs/alloy/issues/3331, once https://github.com/alloy-rs/alloy/pull/3344 is released.

## Solution

Split previous `TransactionBuilder` impl, and move it to rpc-types

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,7 +85,7 @@ jobs:
           cache-on-failure: true
       - name: cargo hack
         run: |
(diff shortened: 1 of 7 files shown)


emhane (commented):
tbh I think it makes most sense to just leave the `TransactionBuilder` impl where it is, and that way don't change the dependencies, just update the impl to remove the generic for `TransactionBuilder` in accordance with pending alloy release

emhane:
marking as draft, when the deps don't point to a git revision we can mark it as ready for review

PR #623, opened 2026-02-03
chore: bump alloy to 1.6.0

gakonst:
Bumps alloy dependencies from 1.1.2 to 1.6.0.

### Changes
- Updates all alloy crates to 1.6.0
- Adds the new `block_timestamp` field to `Transaction` struct initializers (added in [alloy#3606](https://github.com/alloy-rs/alloy/pull/3606))


mattsse (changes requested):
@gakonst dont do the version update but make it so that this changes becomes non breaking by using ..Default::default where yu added block_timestamp and allow the clippy lint

